PR #45, opened 2023-11-19
Added support for YARN scaling

tgaddair:
Following implementation from https://github.com/jquesnelle/yarn/blob/master/scaled_rope/LlamaYaRNScaledRotaryEmbedding.py

Example model: https://huggingface.co/NousResearch/Yarn-Mistral-7b-128k


geoffreyangus (approved):
LGTM – do we need to update docs anywhere?

PR #47, opened 2023-11-21
Fixed tensor parallelism splits

### LINKED ISSUE #46: Sharded adapters not working
### System Info

Model info:

```
{
  "model_id": "mistralai/Mistral-7B-Instruct-v0.1",
  "model_sha": "7ad5799710574ba1c1d953eba3077af582f3a773",
  "model_dtype": "torch.float16",
  "model_device_type": "cuda",
  "model_pipeline_tag": "text-generation",
  "max_concurrent_requests": 128,
  "max_best_of": 2,
  "max_stop_sequences": 4,
  "max_input_length": 1024,
  "max_total_tokens": 2048,
  "waiting_served_ratio": 1.2,
  "max_batch_total_tokens": 1102544,
  "max_waiting_tokens": 20,
  "validation_workers": 2,
  "version": "0.1.0",
  "sha": null,
  "docker_label": null
}
```

2 A100 gpus, `NVIDIA-SMI 535.104.05             Driver Version: 535.104.05   CUDA Version: 12.2` outside docker. 

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [ ] My own modifications

### Reproduction

Run mistral example with docker on 2 gpus: 

```
model=mistralai/Mistral-7B-Instruct-v0.1
volume=$PWD/data # share a volume with the Docker container to avoid downloading weights every run

docker run --gpus all --shm-size 1g -p 8080:80 -v $volume:/data ghcr.io/predibase/lorax:latest --model-id $model --num-shard 2
```

Then try to generate:

```
❯ curl 127.0.0.1:8080/generate \
    -X POST \
    -d '{"inputs": "[INST] Natalia sold clips to 48 of her friends in April, and then she sold half as many clips in May. How many clips did Natalia sell altogether in April and May? [/INST]", "parameters": {"max_new_tokens": 64, "adapter_id": "vineetsharma/qlora-adapter-Mistral-7B-Instruct-v0.1-gsm8k"}}' \
{"error":"Request failed during generation: Server error: local variable 'lora_b' referenced before assignment","error_type":"generation"}%
```

Basically the issue that when trying to multiply first lora_a matrix, we get it sharded with shape `[2048, r]` while input is not sharded and has shape `[49, 4096]` . 

### Expected behavior

Generation completed successfully

tgaddair:
After removing the layer abstraction over the LoRA weights, we introduced a transpose operation, which meant we needed to be splitting on dim=1 rather than dim=0. This was causing tensor parallelism to break, affecting all deployments with more than one GPU.

This PR also fixes support for o_proj, which is row-parallel and needs to be split on dim=0. Previously, there was a bug preventing k_proj and o_proj from being picked up correctly, which is why this was missed.

Closes #46.

### DIFF
--- a/server/Makefile
+++ b/server/Makefile
@@ -24,8 +24,8 @@ install: gen-server install-torch
 	pip install -e ".[bnb, accelerate]"
 
 run-dev:
-	# SAFETENSORS_FAST_GPU=1 python -m torch.distributed.run --nproc_per_node=1 lorax_server/cli.py serve meta-llama/Llama-2-7b-hf --sharded
-	SAFETENSORS_FAST_GPU=1 python -m torch.distributed.run --nproc_per_node=2 lorax_server/cli.py serve mistralai/Mistral-7B-Instruct-v0.1 --sharded
+	SAFETENSORS_FAST_GPU=1 python -m torch.distributed.run --nproc_per_node=2 lorax_server/cli.py serve meta-llama/Llama-2-7b-hf --sharded
+	# SAFETENSORS_FAST_GPU=1 python -m torch.distributed.run --nproc_per_node=2 lorax_server/cli.py serve mistralai/Mistral-7B-Instruct-v0.1 --sharded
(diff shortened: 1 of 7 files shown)


geoffreyangus (approved):
LGTM

PR #54, opened 2023-11-22
Add support for paged attention v2 and update flash attention v2

tgaddair:


server/lorax_server/models/custom_modeling/flash_neox_modeling.py
```diff
@@ -27,11 +27,8 @@
 from transformers.models.gpt_neox import GPTNeoXConfig
 from typing import Optional, List, Tuple
 
-# vllm imports
-import vllm_cache_ops
-import vllm_attention_ops
-
 from lorax_server.utils.flash_attn import attention
+from lorax_server.utils import paged_attn
```
flozi00, line 31: simple design question: why not using flash_attn prefix as with paged_attn
so it would be easier to understand that flash attention is used
if its only attention there could be missunderstandings which one it is

tgaddair: Good point! I'll update so it will be called as `flash_attn.attention` instead.

tgaddair: Updated.

flozi00: LGTM

PR #58, opened 2023-11-22
Fixed adapter loading for GPTQ base models

### LINKED ISSUE #56: Merging non gptq adapter to gptq model
### System Info

master branch

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [ ] My own modifications

### Reproduction

start launcher with gptq model, then try to load non gptq adapter

`2023-11-22T19:34:06.185127Z ERROR lorax_client: router/client/src/lib.rs:33: Server error: 'QuantLinear' object has no attribute 'weight'`

for tests if commented out the ID check
```

#if adapter_config.base_model_name_or_path != model_id:
    #    raise ValueError(f"Adapter '{adapter_id}' is not compatible with model '{model_id}'. "
    #                        f"Use --model-id '{adapter_config.base_model_name_or_path}' instead.")
```

I am already thinking about better check, because it depends on model arch and parameters count instead of the id directly. so zephyr lora could be merged successfully to mistral instruct too.

### Expected behavior

detect gptq model and use qweight instead of weight when merging

tgaddair:
Fixes #56.

### DIFF
--- a/server/lorax_server/utils/adapter.py
+++ b/server/lorax_server/utils/adapter.py
@@ -3,6 +3,7 @@ from collections import defaultdict
 from functools import lru_cache
 from pathlib import Path
 from typing import List, Dict, Set, Tuple
+import warnings
 
 import torch
 from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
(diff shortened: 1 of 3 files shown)


server/lorax_server/utils/adapter.py
```diff
@@ -27,8 +28,10 @@ def load_module_map(model_id, adapter_id, adapter_source, weight_names):
     config_path = get_config_path(adapter_id, adapter_source)
     adapter_config = LoraConfig.from_pretrained(config_path)
     if adapter_config.base_model_name_or_path != model_id:
```
flozi00, line 30: Would it make sense to download the config of the base model behind adapter and compare with the base model served ? If it's the same model arch just warn and if the arch is different raise an error ?
Could be problematic when someone tries to load bert adapter to llama

server/lorax_server/utils/gptq/quant_linear.py
```diff
@@ -357,3 +357,7 @@ def forward(self, x):
         )
         out = out + self.bias if self.bias is not None else out
         return out.reshape(out_shape)
+    
+    @property
```
flozi00, line 361: LGTM

server/lorax_server/utils/adapter.py
tgaddair: Good point! Let me add that.

flozi00:
LGTM
Had no chance to run it, only reading the diff.
Left an idea for the warning part.

Thanks for this fast patch :)

server/lorax_server/utils/adapter.py
tgaddair: Updated.

flozi00:
Pretty cool 😎

PR #63, opened 2023-11-26
Vendor punica kernels

tgaddair:
A follow-up PR will make some changes to the existing SGMV kernel from [Punica](https://github.com/punica-ai/punica). In order to make it easier to track and manage changes to these kernels, this PR will vendor in the kernels into the repo instead of maintaining them within a fork of the upstream Punica repo.


abidwael (approved):
LGTM, with minor comment. Also assuming that everything under `server/punica_kernels/` is coming from the submodule.

PR #69, opened 2023-11-27
Add DISABLE_SGMV env var to explicitly fallback to loop

tgaddair:
SGMV kernel is still somewhat experimental. There are some cases (particular ranks, etc.) that can result in erroneous output. Until these issues are fully resolved, this escape hatch will allow users to fallback to the slower but correct loop implementation.

Example:

```
DISABLE_SGMV=1
```

### DIFF
--- a/server/lorax_server/utils/layers.py
+++ b/server/lorax_server/utils/layers.py
@@ -20,7 +20,7 @@ from accelerate import init_empty_weights
 
 try:
     from lorax_server.utils.sgmv import add_lora_sgmv_cutlass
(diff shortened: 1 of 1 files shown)


server/lorax_server/utils/layers.py
```diff
@@ -20,7 +20,7 @@
 
 try:
     from lorax_server.utils.sgmv import add_lora_sgmv_cutlass
-    HAS_SGMV = True
+    HAS_SGMV = not bool(os.environ.get("DISABLE_SGMV", ""))
```
flozi00, line 23: LGTM

flozi00 (approved):

PR #82, opened 2023-11-29
fix: Assign dtype of lora to base model dtype

### LINKED ISSUE #74: Quantized models fail to generate expected output
Example:

```
mistralai/Mistral-7B-v0.1 --quantize bitsandbytes-nf4
```

Request:

```
    -X POST \
    -d '{"inputs": "<|system|> You are a helpful assistant <|user|> What is deep learning? </s> <|assistant|>", "parameters": {"max_new_tokens": 64, "adapter_id": "qblocks/mistral_7b_norobots"}}' \
```

Response:

```
{"generated_text":""}
```

Expected:

```
{"generated_text":"Deep learning is a subset of machine learning that uses artificial neural networks to learn from data. It is a powerful tool for solving complex problems in fields such as natural language processing, computer vision, and speech recognition. Deep learning algorithms can learn from large amounts of data and make predictions or decisions based on that data. They can"}
```

tgaddair:
Fixes #74.

When the model is quantized, the dtype of the weights (compressed) is typically uint or similar. This means if we cast the lora weights to the weight of the parent layer, it will corrupt them, rendering the output garbage. Instead, we should assign the dtype of the base model (float16 or bloat16) to the lora weights, which is consistent with how inference in QLoRA is done outside LoRAX (quantized weights + unquantized lora).

### DIFF
--- a/server/lorax_server/models/flash_causal_lm.py
+++ b/server/lorax_server/models/flash_causal_lm.py
@@ -771,8 +771,8 @@ class FlashCausalLM(Model):
                 # There is no LoRA weight for this layer type in the adapter
                 return
             
-            lora_a = module_map[weight_name]["lora_A"].to(base_device, base_weight.dtype)
-            lora_b = module_map[weight_name]["lora_B"].to(base_device, base_weight.dtype)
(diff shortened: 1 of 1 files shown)


geoffreyangus (approved):
wow, nice catch!

PR #89, opened 2023-11-30
decrease the max batch total tokens manualy

flozi00:
this is a first dry try ;)
I did not run this and never used rust before


tgaddair (approved):

flozi00:
🙏 thank you for the fix
Ready to be merged from my side

PR #111, opened 2023-12-07
add support for H100s

thelinuxkid:
# What does this PR do?

add support for H100s

## Before submitting
- [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Was this discussed/approved via a Github issue or the discord / slack channel? Please add a link
      to it if that's the case.
- [x] Did you write any new necessary tests?

## Who can review?

@arnavgarg1 
@tgaddair 

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

 
 ## POC
image

PR #118, opened 2023-12-09
Add h100 support

tgaddair:
Continuation of #111.

Pin urllib3 due to https://github.com/python-poetry/poetry-plugin-export/issues/183


tgaddair:
- [ ]

PR #121, opened 2023-12-11
upgrade to py 3.10

flozi00:
# What does this PR do?
replacing all the python specific version numbers to 3.10 to be ready for integrating the mixtral kernels which needs python 3.10+


tgaddair (approved):
Ran tests and verified everything works as intended!

PR #125, opened 2023-12-13
Add predibase as a source for adapters

magdyksaleh:
This PR adds predibase as a source for adapters.

It implements a thin wrapper over the S3 model source and resolves the model name and version to a s3 bucket path to pull the adapters from

- [ ] ~~Move token to header so it isn't logged~~ 
- [ ] ~~decouple bucket~~
- [x] make version optional 
- [ ] ~~Rename to token from predibase token~~


launcher/src/main.rs
```diff
@@ -100,9 +100,10 @@ struct Args {
     source: String,
 
     /// The source of the model to load.
-    /// Can be `hub` or `s3`.
+    /// Can be `hub` or `s3` or 'predibase'
```
magdyksaleh, line 103: Will change to pbase for consistency

magdyksaleh:
Todo: move the api token to be read from the request so different tenants can have different tokens for the same endpoint

server/lorax_server/utils/sources/__init__.py
```diff
@@ -7,6 +11,21 @@
...
+
+@lru_cache(maxsize=256)
+def map_pbase_model_id_to_s3(model_id: str) -> str:
+    name, version = model_id.split("/")
+    url = PREDIBASE_GATEWAY_ENDPOINT + PREDIBASE_MODEL_URL_ENDPOINT.format(name, version)
+    headers = {"Authorization": f"Bearer {PREDIBASE_API_TOKEN}"}
+    resp = requests.get(url, headers=headers)
+    resp.raise_for_status()
+    uuid, best_run_id = resp.json()["uuid"], resp.json()["bestRunID"]
+    return f"{uuid}/{best_run_id}/artifacts/model/model_weights/"
```
tgaddair, line 28: This feels a little too coupled to backend details. Can we have an endpoint that returns the full s3 URI: `s3://bucket/.../model_weights/`? I realize the current s3 source is coupled to an environment variable that specifies the bucket, but ideally we should change that, too (it makes it less useful for OSS users who want to use s3).

server/lorax_server/utils/sources/__init__.py
```diff
@@ -7,6 +11,21 @@
...
+
+PREDIBASE_API_TOKEN = os.getenv("PREDIBASE_API_TOKEN", None)
+PREDIBASE_MODEL_URL_ENDPOINT = "/v1/models/version/name/{}?version={}"
+PREDIBASE_GATEWAY_ENDPOINT = os.getenv("PREDIBASE_GATEWAY_ENDPOINT", "https://api.predibase.com")
+
+@lru_cache(maxsize=256)
+def map_pbase_model_id_to_s3(model_id: str) -> str:
+    name, version = model_id.split("/")
+    url = PREDIBASE_GATEWAY_ENDPOINT + PREDIBASE_MODEL_URL_ENDPOINT.format(name, version)
+    headers = {"Authorization": f"Bearer {PREDIBASE_API_TOKEN}"}
```
tgaddair, line 24: This works for dedicated deployments that are single-tenant, but would not work for serverless deployments, where each request could have a different API token. I think it's fine if we implement it like this for now, but we would need to revisit for serverless LLMs soon.

server/lorax_server/utils/sources/__init__.py
```diff
@@ -7,6 +11,21 @@
...
+PBASE = "pbase"
+
+PREDIBASE_API_TOKEN = os.getenv("PREDIBASE_API_TOKEN", None)
+PREDIBASE_MODEL_URL_ENDPOINT = "/v1/models/version/name/{}?version={}"
+PREDIBASE_GATEWAY_ENDPOINT = os.getenv("PREDIBASE_GATEWAY_ENDPOINT", "https://api.predibase.com")
+
+@lru_cache(maxsize=256)
+def map_pbase_model_id_to_s3(model_id: str) -> str:
+    name, version = model_id.split("/")
+    url = PREDIBASE_GATEWAY_ENDPOINT + PREDIBASE_MODEL_URL_ENDPOINT.format(name, version)
```
tgaddair, line 23: Can we allow the version to be optional?

server/lorax_server/utils/sources/__init__.py
magdyksaleh: Good point! Will fix now

server/lorax_server/utils/sources/__init__.py
magdyksaleh: yep have a todo to add the api_token as a header in the incoming request

router/src/infer.rs
```diff
@@ -138,7 +138,8 @@ impl Infer {
             }
         }
 
-        let adapter = Adapter::new(adapter_id.unwrap(), adapter_source.unwrap(), adapter_idx);
+        let predibase_api_token = request.parameters.predibase_api_token.clone();
```
tgaddair, line 141: Can we get this from the bearer token in the header so the user dosn't need to put it in the request body?

proto/generate.proto
```diff
@@ -224,13 +224,17 @@ enum AdapterSource {
...
+    PBASE = 3;
 }
 
 message DownloadAdapterRequest {
     /// Adapter ID
     string adapter_id = 1;
     /// Adapter source
     AdapterSource adapter_source = 2;
+    /// API Token for predibase
+    optional string predibase_api_token = 3;
```
tgaddair, line 237: Can we rename to just `api_token`? I could see this being useful for HF hub token as well.

router/src/infer.rs
tgaddair: Actually, on second thought, that wouldn't work well for HF per my other comment, so I'm fine keeping this as a param.

server/lorax_server/query.py
```diff
@@ -0,0 +1,17 @@
+import requests
```
tgaddair, line 1: Looks like this was added by mistake?

tgaddair (commented):
Hey @magdyksaleh, I see you requested my review on this, but it looks like my two comments were not addressed yet.

server/lorax_server/query.py
magdyksaleh: yep a test file

proto/generate.proto
magdyksaleh: ok will rename

magdyksaleh:
sorry my comment wasn't added

server/lorax_server/utils/sources/__init__.py
tgaddair: Okay, it's not a release blocker, but it is exposing a lot of unnecessary detail about our underlying storage system that will break if we ever move files around or want to deploy in Azure, etc. So definitely something we need to revisit.

server/lorax_server/utils/sources/__init__.py
```diff
@@ -7,6 +11,30 @@
...
 LOCAL = "local"
+PBASE = "pbase"
+
+PREDIBASE_MODEL_URL_ENDPOINT = "/v1/models/version/name/{}"
+PREDIBASE_MODEL_VERSION_URL_ENDPOINT = "/v1/models/version/name/{}?version={}"
+PREDIBASE_GATEWAY_ENDPOINT = os.getenv("PREDIBASE_GATEWAY_ENDPOINT", "https://api.predibase.com")
+
+
+@lru_cache(maxsize=256)
+def map_pbase_model_id_to_s3(model_id: str, api_token: str) -> str:
```
tgaddair, line 22: API token is optional, right? We should raise a useful error if the user didn't provide it.

tgaddair (commented):
Looks like there are compiler errors.

tgaddair (approved):
LGTM!

PR #126, opened 2023-12-13
Update client docs with new endpoint source

abidwael:


docs/reference/python_client.md
```diff
@@ -67,7 +67,7 @@ Example:
 ```python
 from lorax import Client
 
-endpoint_url = f"https://api.app.predibase.com/v1/llms/{llm_deployment_name}"
+endpoint_url = f"https://serving.app.predibase.com/{predibase_tenant_id}/llm/llm-deployment-{llm_deployment_name}/generate"
```
brightsparc, line 70: Can we remove the `llm-deployment-` when resolving the URL in the backend - we would need to update the VS for this too.

tgaddair: +1, less verbose the better.

tgaddair: Can you also explain how to obtain the tenant ID?

tgaddair: Also, we need to use `/` endpoint, not `/generate`.

abidwael: updated with information on the tenant short code. I'll need to update the VS to add `/` as a supported route

brightsparc (approved):

docs/reference/python_client.md
```diff
@@ -67,7 +67,9 @@ Example:
 ```python
 from lorax import Client
 
-endpoint_url = f"https://api.app.predibase.com/v1/llms/{llm_deployment_name}"
+# You can get your Predibase API token by going to Settings > My Profile > Generate API Token
+# You can get your Predibase Tenant short code by going to Settings > My Profile > Overview > Tenant ID
+endpoint_url = f"https://serving.app.predibase.com/{predibase_tenant_short_code}/deployments/v2/llms/{llm_deployment_name}/generate"
```
tgaddair, line 72: This won't work with streaming endpoint, though. It really should be `/`, not `/generate`.

clients/python/lorax/client.py
```diff
@@ -155,7 +155,9 @@ def generate(
         payload = resp.json()
         if resp.status_code != 200:
             raise parse_error(resp.status_code, payload)
-        return Response(**payload[0])
+        if isinstance(payload, list):
+            return Response(**payload[0])
```
tgaddair, line 159: This smells like bad design IMO. The REST API should be consistent and predictable. We should not be writing client-side hacks to account for inconsistency in the REST API.

PR #132, opened 2023-12-15
Added Phi

tgaddair:
Example:

```shell
lorax-launcher --model-id microsoft/phi-2
```

Prompt:

```shell
    -X POST \
    -d '{"inputs": "Instruct: Write a detailed analogy between mathematics and a lighthouse.\nOutput:", "parameters": {"max_new_tokens": 64}}' \
```

Response:

```
Mathematics is like a lighthouse. Just as a lighthouse guides ships safely to shore, mathematics provides a guiding light in the world of numbers and logic. It helps us navigate through complex problems and find solutions. Just as a lighthouse emits a steady beam of light, mathematics provides a consistent framework for reasoning and problem-solving.
```


server/lorax_server/models/__init__.py
```diff
@@ -325,6 +328,19 @@ def get_model(
                 trust_remote_code=trust_remote_code,
             )
         raise NotImplementedError("Qwen model requires flash attention v2")
+    
+    if model_type == "phi-msft" or model_type == "phi":
```
flozi00, line 332: what do you think about using model_type in [] instead ?

flozi00 (approved):

server/lorax_server/models/custom_modeling/flash_phi_modeling.py
```diff
@@ -0,0 +1,396 @@
...
+        self,
+        prefix: str,
+        config,
+        weights,
+        layer_id: int,
+    ):
+        super().__init__()
+        self.num_heads = config.n_head
+        self.hidden_size = config.n_embd
+        self.head_size = self.hidden_size // self.num_heads
```
arnavgarg1, line 78: Is it intentionally to use `self.num_heads` like this here but then update the value of `self.num_heads` by the number of process groups in line 110?

server/lorax_server/models/flash_phi.py
```diff
@@ -0,0 +1,178 @@
...
+            dtype = torch.float16 if dtype is None else dtype
+        else:
+            raise NotImplementedError("FlashPhi is only available on GPU")
+
+        tokenizer = AutoTokenizer.from_pretrained(
+            model_id,
+            revision=revision,
+            padding_side="left",
+            truncation_side="left",
+            trust_remote_code=trust_remote_code,
```
arnavgarg1, line 58: Since we're adding `trust_remote_code`, do we already install the `einops` library to convert weights when using `microsoft/phi-1.5`?

arnavgarg1 (approved):

server/lorax_server/models/flash_phi.py
tgaddair: `trust_remote_code` is `False` by default. `einops` is installed already (for Falcon). This current implementation works with the microsoft version of the weights, rather than the changes HF made.

server/lorax_server/models/custom_modeling/flash_phi_modeling.py
tgaddair: Yes, when creating the modules we need the `num_heads` value pre-splitting. Then later on we use `num_heads` post-split.

server/lorax_server/models/__init__.py
tgaddair: Done.

server/lorax_server/models/custom_modeling/flash_phi_modeling.py
tgaddair: Added comment.

PR #138, opened 2023-12-18
speculative decoding

flozi00:
# What does this PR do?
Still WIP
At the moment i mostly copied the tokenchoser and medusa model from tgi, need to refactor that due to license issues

opened the PR for progress tracking or if you should have motivation to overtake that @tgaddair ;-)


flozi00:
having problems with docker again
was not able to test the implementation

tgaddair:
@flozi00 I can try building a docker image if you'd like. Would that help with testing?

flozi00:
I have some questions to discuss about the use case and implementation.
Since medusa is adding more heads i thought it was compatible with lora, but now i am thinking about if it should be used instead of lora.
model.lm_head will be set to multiple heads
when calling lora id it tries to use the lm_head from adapter

PR #146, opened 2023-12-21
Update client docs with new endpoint source

abidwael:


docs/reference/python_client.md
```diff
@@ -67,7 +67,9 @@ Example:
 ```python
 from lorax import Client
 
-endpoint_url = f"https://api.app.predibase.com/v1/llms/{llm_deployment_name}"
+# You can get your Predibase API token by going to Settings > My Profile > Generate API Token
+# You can get your Predibase Tenant short code by going to Settings > My Profile > Overview > Tenant ID
+endpoint_url = f"https://serving.app.predibase.com/{predibase_tenant_short_code}/deployments/v2/llms/{llm_deployment_name}/"
```
tgaddair, line 72: Do we need trailing slash?

tgaddair (approved):

tgaddair (approved):

docs/reference/python_client.md
abidwael: we can with or without, but we no longer need it

abidwael: for record-keeping, using `/generate` will not work in the client because the API will not return a list.

PR #157, opened 2024-01-04
remove CACHE_MANAGER in flash_causal_lm.py

michaelfeil:
global var CACHE_MANAGER should no longer be needed here, fyi.


tgaddair (approved):
Thanks for the PR!

PR #168, opened 2024-01-09
fix: Handle NaN values during weight conversion

### LINKED ISSUE #81: Surface more informative error when adapter has NaN weights
### Feature request

When querying a base model with an adapter that has NaN or Inf weight tensors, LoRAX returns the following error:

```
The output tensors do not match for key base_model.model.model.layers.0.self_attn.q_proj.lora_A.weight
```

It would be more helpful if the error message indicates the the reason the tensors don't match during merge is because LoRAX detected NaN/Inf tensors in the adapter weights. 

### Motivation

This would help provide a rectifiable/actionable path for users who fine-tuned models and are working on testing them out to know that this isn't an issue with LoRAX, but rather, an issue with their trained adapter weights.

### Your contribution

Happy to help surface a better error message! Seems like the issue is raised from this line in particule? https://github.com/predibase/lorax/blob/360ad4c8f903bdcf129e0e90a97d645ae5a5c2f3/server/lorax_server/utils/convert.py#L92

tgaddair:
Fixes #81.

### DIFF
--- a/server/lorax_server/utils/convert.py
+++ b/server/lorax_server/utils/convert.py
@@ -8,6 +8,8 @@ from safetensors.torch import save_file, load_file, _find_shared_tensors, _is_co
 from typing import List, Dict
 from collections import defaultdict
 
+from lorax_server.utils.errors import NanWeightsError
+
 
 def _remove_duplicate_names(
     state_dict: Dict[str, torch.Tensor],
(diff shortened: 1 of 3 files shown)


arnavgarg1 (approved):
Nice! LGTM, thanks for putting this together @tgaddair! Is it also worth adding special error checking if the weights have Inf (I know this is super rare and really only can happen when training without regularization measures in place).

asingh9530:
@tgaddair looks good but not sure if nesting conditions would be right choice as if for any reason we in future it requires to add more check it would start becoming clustered.

tgaddair:
@asingh9530 it's a good point. My reason for doing it this way was to save the computation cost of calling `torch.any(torch.isnan))` in cases where we know the weights are not nan. Not sure how much of an impact this will have in practice. Happy to go with an alternative if you feel there's a better place for this check!

asingh9530:
@tgaddair how about using a  `validatorTensor` method and it takes care of validating tensors and then we can add validation conditions much neatly there. Let me know I can add in this PR as adding  internal methods like` validation` or `convert` makes a huge difference in maintaining

tgaddair:
Hey @asingh9530, that sounds good! I can go ahead and merge this PR for now and then you can create a follow-up PR, if that works.

PR #201, opened 2024-01-23
Add a model cache to avoid running out of storage

magdyksaleh:


sync.sh
```diff
@@ -20,6 +25,75 @@ is_launcher_running() {
...
+    # Calculate number of lines to be removed, if any
+    local lines_to_remove=$((total_lines - CACHE_SIZE))
+
+    if [ "$lines_to_remove" -gt 0 ]; then
+        # Store removed lines in a variable
+        removed_lines=$(head -n "$lines_to_remove" "$temp_file")
+        echo "Deleting $removed_lines from cache"
+    fi
+
+    # Ensure only the last 5 items are retained
```
brightsparc, line 61: Nit, "5" should be N given its base don env var.

brightsparc (approved):
All done in the shell, impressive.

Given this isn't based on size of cache, instead a number, do we have an estimate of how many GB 4 could be - it seems a little on the low side?

Will you have a seperate method to append to the CACHE_FILE, we will want to make sure this file doesn't contain duplicate models too.

magdyksaleh:
> All done in the shell, impressive.
> 
> Given this isn't based on size of cache, instead a number, do we have an estimate of how many GB 4 could be - it seems a little on the low side?
> 
> Will you have a seperate method to append to the CACHE_FILE, we will want to make sure this file doesn't contain duplicate models too.

If the model is already in the file the grep command removes it and then we append the model name to the bottom of the file. That way the file is always ordered from least to most recently used. When we remove lines off the top an delete those directories those are the least recently used models.

tgaddair (approved):
@magdyksaleh LGTM. Please feel free to land when you're ready.

PR #207, opened 2024-01-24
Added list of exported metrics

noyoshi:
Added a list of the exported metrics to the readme. 

Further info would be nice to add to the table - such as the metric type and the description.


jppgks (approved):

README.md
```diff
@@ -163,13 +162,56 @@ print("Response:", resp[0].choices[0].message.content)
 
 See [OpenAI Compatible API](https://predibase.github.io/lorax/guides/openai_api) for details.
 
+### Metrics
```
tgaddair, line 165: Can you add this to docs instead? We have a section called Reference, we could add a page called Prometheus Metrics or similar.

tgaddair: Also, is this something we can expose via the `openapi.json`?

PR #220, opened 2024-02-02
Add model and adapter headings

magdyksaleh:


router/src/server.rs
```diff
@@ -397,6 +404,15 @@ async fn generate(
         time_per_token.as_millis().to_string().parse().unwrap(),
     );
 
+    headers.insert(
+        "x-predibase-model-id",
```
jeffreyftang, line 408: Nit: Might be a bit strange to hard-code predibase-specific things into an open-source library.

brightsparc: Agree. Can we just return the base model name / size?

magdyksaleh: sure can just make it model-id and adapter-id

brightsparc (approved):
This would suffice for immediate needs, but how would this work for HF hosted models not loaded from PB?

router/src/server.rs
brightsparc: Convection is still to use X for custom headers so would something like `x-lorax-model=pb://model-name` make sense and cater for HF models too?

magdyksaleh:
> This would suffice for immediate needs, but how would this work for HF hosted models not loaded from PB?

This is agnostic to model source but we can also make sure we pass a model source header so that you can infer the pb internals

router/src/server.rs
```diff
@@ -286,6 +292,7 @@ async fn generate(
     }
 
     let details = req.0.parameters.details || req.0.parameters.decoder_input_details;
+    let adapter_id = req.0.parameters.adapter_id.clone();
```
tgaddair, line 295: Note that following #212, adapter ID can come from the `merged_parameters` as well as the `adapter_id`.

router/src/server.rs
```diff
@@ -31,6 +33,10 @@ use tracing::{info_span, instrument, Instrument};
 use utoipa::OpenApi;
 use utoipa_swagger_ui::SwaggerUi;
 
+lazy_static! {
+    static ref MODEL_ID: Mutex<String> = Mutex::new("".to_string());
```
tgaddair, line 37: What does this do?

tgaddair: It smells like we're setting a global lock on the current model ID, which would break concurrency?

magdyksaleh: It allows you to create a global var that we then update [here](https://github.com/predibase/lorax/pull/220/files#diff-58724fd9e4e3f359bc071bec55c6b3f8149372efa3dc33981ada5bf60afee878R844-R847) with the model name. Needed so it is not read only.

router/src/server.rs
magdyksaleh: good point

router/src/server.rs
tgaddair: But why do we need it? I see we set it, but don't see where it gets used.

tgaddair: I see, but why do we need to lock every time we read it if it's only initialized once at the beginning and then read-only? Feels like we're taking a concurrency hit for no reason.

magdyksaleh: @tgaddair we just set it once when we run the server for the first time. Maybe we can have a read-write lock. Or maybe we can do something different but we want to communicate the value of model_id [here](https://github.com/predibase/lorax/pull/220/files#diff-58724fd9e4e3f359bc071bec55c6b3f8149372efa3dc33981ada5bf60afee878R768) to the request handler [here](https://github.com/predibase/lorax/pull/220/files#diff-58724fd9e4e3f359bc071bec55c6b3f8149372efa3dc33981ada5bf60afee878R855). Maybe we can create a partial function like in python? idk

magdyksaleh: @tgaddair I think we would need to lock a global var but I could be wrong here

magdyksaleh: Actually looks like I can just use this https://docs.rs/once_cell/latest/once_cell/

tgaddair: Looks like exactly what we want! Nice find.

router/src/infer.rs
```diff
@@ -123,24 +123,11 @@ impl Infer {
...
-        let adapter_parameters =
-            request
-                .parameters
-                .adapter_parameters
-                .clone()
-                .unwrap_or(AdapterParameters {
-                    adapter_ids: vec![adapter_id.clone().unwrap()],
-                    ..Default::default()
-                });
+        let (adapter_id, adapter_source, adapter_parameters) = extract_adapter_params(
```
tgaddair, line 126: I think we may want to just discard adapter_id at this point, as it should b contained in the adapter parameters.

So I would imagine this function just returning `adapter_source` and `adapter_parameters`.

tgaddair (approved):

PR #221, opened 2024-02-05
Add Cors params

magdyksaleh:


tgaddair (approved):

brightsparc:
Would be good to also update the [docs](https://github.com/predibase/lorax/blob/main/docs/reference/launcher.md#L217) to include this.

PR #234, opened 2024-02-10
Update s3.py

llama-shepard:
# What does this PR do?
This PR is a fix for lorax.errors.GenerationError: Request failed during generation: Server error: Can't find 'adapter_config.json' at '/data/models--s3:----adapters--gsm8k/snapshots'

The adapter is getting downloaded to models--{model_id}/snapshots folder. But get_local_path is returning path "/data/models--s3:----{bucket}--{model_id}/snapshots". 

This fix returns the "models--{model_id}/snapshots" path

## Before submitting
      to it if that's the case.

## Who can review?
@tgaddair


tgaddair (approved):
Good catch! This is definitely an oversight when we introduced the `_get_bucket_and_model_id` helper.

CI is failing due to a Rust dep, which is a known issue, so I'll land this as is for now.

PR #265, opened 2024-02-21
Pass details param into client 

magdyksaleh:
Currently we pass details=True for all client generate calls - This relaxes that constraint.


jeffreyftang (approved):

tgaddair (commented):
Should we just call the param `details` to be consistent with the REST API?

magdyksaleh:
I was thinking about that but details is just such a bad name haha - I want to rename it in the REST API so it is clear. What do you think @tgaddair ?

tgaddair:
@magdyksaleh agree it's not a great name, but would err on the side of stability for now.

PR #273, opened 2024-02-22
fix: Remove vec wrapping from OpenAI-compatible response

### LINKED ISSUE #272: The OpenAI API doesn't match the api-reference
### System Info

latest version

### Information

- [ ] Docker
- [ ] The CLI directly

### Tasks

- [ ] My own modifications

### Reproduction

- Start LoraX latest version
- Send request to LoRaX server via OpenAI API

### Expected behavior

In this line: https://github.com/predibase/lorax/blob/a3b865d8b033db7602f485491ab4091f1e5b2ea6/router/src/server.rs#L203
The server convert the response to list but the [OpenAI API reference](https://platform.openai.com/docs/api-reference/making-requests) show that the response is just a single object:
image

That issue raises some problems when I use some libraries related to OpenAI like LiteLLM,...
Can you help me update?

jeffreyftang:
Closes #272

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2580,6 +2580,7 @@ version = "1.0.102"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b5062a995d481b2308b6064e9af76011f2921c35f97b0468811ed9f6cd91dfed"
 dependencies = [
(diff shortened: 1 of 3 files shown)


tgaddair:
Just need to fix Rust format

PR #278, opened 2024-02-23
enh: Propagate bearer token from header if one exists for OpenAI-compatible endpoints

jeffreyftang:
Since the OpenAI client doesn't include an API token or authentication key as part of the request payload, it's useful to propagate the bearer token from the incoming request headers so that downstream calls can use it to authenticate.

Alternatively, we could support passing such a token more explicitly via an additional header, which the OpenAI client [does support](https://github.com/openai/openai-python/blob/a7115b5f33acd27326e5f78e19beb0d73bd3268e/src/openai/resources/chat/completions.py#L88). This is more expressive, but would require the user to take more action.

### DIFF
--- a/router/src/server.rs
+++ b/router/src/server.rs
@@ -64,6 +64,7 @@ example = json ! ({"error": "Incomplete generation"})),
 async fn compat_generate(
     default_return_full_text: Extension<bool>,
     infer: Extension<Infer>,
+    req_headers: HeaderMap,
     req: Json<CompatGenerateRequest>,
 ) -> Result<Response, (StatusCode, Json<ErrorResponse>)> {
     let mut req = req.0;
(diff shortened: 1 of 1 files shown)


router/src/server.rs
```diff
@@ -111,11 +111,21 @@ example = json ! ({"error": "Incomplete generation"})),
...
     default_return_full_text: Extension<bool>,
     infer: Extension<Infer>,
+    headers: HeaderMap,
     req: Json<CompletionRequest>,
 ) -> Result<Response, (StatusCode, Json<ErrorResponse>)> {
     let req = req.0;
     let mut gen_req = CompatGenerateRequest::from(req);
 
+    // If a bearer token was present in this request, pass the associated token along.
+    let _ = headers.get("authorization").map_or((), |x| {
```
tgaddair, line 121: We should do this for the generate and generate_stream endpoints as well (if the user did not pass a token explicitly).

tgaddair (approved):

router/src/server.rs
jeffreyftang: Done.

PR #282, opened 2024-02-26
add in response headers to streaming endpoint

noyoshi:
- Added response headers from `generate` into `generate_stream` endpoint. Next will be to add them to the open AI compatible endpoints.

### DIFF
--- a/router/src/server.rs
+++ b/router/src/server.rs
@@ -554,6 +554,26 @@ async fn generate_stream_with_callback(
         });
     }
 
+    let (adapter_source, adapter_parameters) = extract_adapter_params(
+        req.0.parameters.adapter_id.clone(),
+        req.0.parameters.adapter_source.clone(),
+        req.0.parameters.adapter_parameters.clone(),
+    );
+
+    let adapter_id_string = adapter_parameters
+        .adapter_ids
+        .iter()
+        .map(|id| id.as_str())
+        // filter out base model adapter id
+        .filter(|id| *id != BASE_MODEL_ADAPTER_ID)
+        .collect::<Vec<_>>()
+        .join(",");
+
+    if adapter_id_string.len() > 0 {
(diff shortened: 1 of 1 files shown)


router/src/server.rs
```diff
@@ -554,6 +554,28 @@ async fn generate_stream_with_callback(
...
+        .filter(|id| *id != BASE_MODEL_ADAPTER_ID)
+        .collect::<Vec<_>>()
+        .join(",");
+
+    if adapter_id_string.len() > 0 {
+        headers.insert("x-adapter-ids", adapter_id_string.parse().unwrap());
+        headers.insert("x-adapter-source", adapter_source.unwrap().parse().unwrap());
+    }
+
+    headers.insert("x-model-id", MODEL_ID.get().unwrap().parse().unwrap());
```
brightsparc, line 577: Maybe we could also include `x-prompt-tokens` in lieu of `x-total-tokens` until we solve log capture?

brightsparc (approved):
Could be useful to add prompt tokens since we should have these when outputting the initial access log?

router/src/server.rs
noyoshi: I think that can wait until another PR

router/src/server.rs
```diff
@@ -554,6 +554,28 @@ async fn generate_stream_with_callback(
...
+        .iter()
+        .map(|id| id.as_str())
+        // filter out base model adapter id
+        .filter(|id| *id != BASE_MODEL_ADAPTER_ID)
+        .collect::<Vec<_>>()
+        .join(",");
+
+    if adapter_id_string.len() > 0 {
+        headers.insert("x-adapter-ids", adapter_id_string.parse().unwrap());
+        headers.insert("x-adapter-source", adapter_source.unwrap().parse().unwrap());
```
jeffreyftang, line 574: Shouldn't there be some guards against panics here?

PR #286, opened 2024-02-28
Generate to `max_total_tokens` during warmup

### LINKED ISSUE #279: LoRAX attempting to serve requests with token length greater than max input tokens
### System Info

Model being used: llama-v2-13b derivative
Hardware used: A10g X 4
Deployment: AWS EC2 cluster
Version: latest 

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [ ] My own modifications

### Reproduction

Lorax seems to successfully attempt serving requests where input tokens is more than the max input tokens parameter set while initializing LoRAX. 

### Expected behavior

Expected behavior is to reject such requests.

tgaddair:
Closes #279.

If the user sets `max_total_tokens` to a value greater than what the model can support, catch the error during warmup (initialization) rather than at request time. This is important because device-side asserts leave the server in a broken state requiring a reset.

### DIFF
--- a/server/lorax_server/models/flash_causal_lm.py
+++ b/server/lorax_server/models/flash_causal_lm.py
@@ -8,6 +8,7 @@ import torch
 import torch.distributed
 from loguru import logger
 from opentelemetry import trace
+from tqdm import tqdm
 from transformers import PreTrainedTokenizerBase
 
 from lorax_server.models import Model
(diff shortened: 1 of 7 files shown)


server/lorax_server/models/flash_causal_lm.py
```diff
@@ -731,7 +732,11 @@ def warmup(self, batch: FlashCausalLMBatch):
                 self.dtype,
                 self.device,
             )
-            _, batch = self.generate_token(batch)
+
+            with tqdm(total=max_new_tokens, desc="Warmup to max_total_tokens") as pbar:
```
jeffreyftang, line 736: Nit: could be nice to put the actual value of `max_total_tokens` in as well.

jeffreyftang (approved):

server/lorax_server/models/flash_causal_lm.py
tgaddair: Good point, though will require a bit more plumbing. We do show the max new tokens from the progress bar counter, so will leave as is for now.

PR #288, opened 2024-02-28
feat: add repetition_penalty and top_k to openai

### LINKED ISSUE #287: Support repetition_penalty in OpenAI API
### Feature request

Support `repetition_penalty` in OpenAI compaitaple API

### Motivation

It's really important parameters, currently not support in OpenAI compaitaple API

### Your contribution

.

huytuong010101:
Fixes #287.

### DIFF
--- a/router/src/lib.rs
+++ b/router/src/lib.rs
@@ -455,6 +455,8 @@ struct ChatCompletionRequest {
     // Additional parameters
     // TODO(travis): add other LoRAX params here
     response_format: Option<ResponseFormat>,
+    repetition_penalty: Option<f32>,
+    top_k: Option<i32>,
 }
 
 #[derive(Clone, Debug, Deserialize, ToSchema)]
@@ -478,6 +480,8 @@ struct CompletionRequest {
     user: Option<String>,
     // Additional parameters
     // TODO(travis): add other LoRAX params here
+    repetition_penalty: Option<f32>,
+    top_k: Option<i32>,
 }
 
 #[derive(Serialize, ToSchema)]
@@ -587,8 +591,8 @@ impl From<CompletionRequest> for CompatGenerateRequest {
                 api_token: None,
                 best_of: req.best_of.map(|x| x as usize),
                 temperature: req.temperature,
-                repetition_penalty: None,
-                top_k: None,
+                repetition_penalty: req.repetition_penalty,
+                top_k: req.top_k,
                 top_p: req.top_p,
                 typical_p: None,
                 do_sample: !req.n.is_none(),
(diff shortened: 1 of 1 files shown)


router/src/lib.rs
```diff
@@ -455,6 +455,8 @@ struct ChatCompletionRequest {
     // Additional parameters
     // TODO(travis): add other LoRAX params here
     response_format: Option<ResponseFormat>,
+    repetition_penalty: Option<f32>,
```
tgaddair, line 458: Looks like the OpenAI spec defines this as `presence_penalty`: https://platform.openai.com/docs/api-reference/chat/create

Since there is no `top_k` equivalent in OpenAI, adding it as a new param is fine.

router/src/lib.rs
```diff
@@ -587,8 +591,8 @@ impl From<CompletionRequest> for CompatGenerateRequest {
                 api_token: None,
                 best_of: req.best_of.map(|x| x as usize),
                 temperature: req.temperature,
-                repetition_penalty: None,
-                top_k: None,
+                repetition_penalty: req.repetition_penalty,
```
tgaddair, line 594: If we map `repetition_penalty` to the OpenAI `presence_penalty`, we need to shift the range from (-2, 2) (OpenAI) to (0, 4) ours, so something like:

```
repetition_penalty: req.presence_penalty.map(|x| x + 2.0)
```

Same for lin 629.

tgaddair (commented):
Thanks for this PR! Just a couple suggestions to better align with OpenAI's format.

router/src/lib.rs
huytuong010101: As I known, presence_penalty and repetition_penalty both have the same effect. But
- presence_penalty - Between -2.0 and 2. value 0 means no penalty. Default to 0.0
- repetition_penalty –  Between 1.0 and infinity. 1.0 means no penalty. Default to 1.0.
So it not have same range, and presence_penalty=0 (no penalty) not have same effect with repetition_penalty=2 (much penalty)
Do you think we need keep it seprately or any better method to shift it ?

prd-tuong-nguyen: I recommend keep both params presence_penalty and repetition_penalty.
Or if you still want to reuse `presence_penalty`, I think `repetition_penalty: req.presence_penalty.map(|x| x + 1.0)` is better.

tgaddair: Hmm, maybe it's fine to keep both for now. We can think about how best to map presence_penalty to repition_penalty in a follow-up. Thanks for the PR!

tgaddair (approved):

router/src/lib.rs
prd-tuong-nguyen: @tgaddair thank u

PR #297, opened 2024-03-04
Add support for returning alternative tokens

### LINKED ISSUE #298: Include logprobs for alternative, less probable tokens in the generation response
### Feature request

For some applications it is necessary to access the full probability distribution of the tokens generated by the model's decoder head. This could be implemented by including a (size-limited) list of alternatives with every generated token in the `details` field of a response. This list would then include the `k` most probable tokens for each generation step, including the logprob and decoded text of the token.

### Motivation

Having access to the full probability distribution of the LLM decoder head allows for a more precise analysis of the model's response. For example, by comparing the logprob of the chosen token with the alternatives, we can deduce how likely the model was to pick a different option for that next token.

### Your contribution

For models based on `FlashCausalLM`, the necessary changes are implemented in #297. Here, the client parameter `return_k_alternatives` can be used to set the size of the list of alternatives that is returned with each generated token. For models based on other classes, setting this option currently has no effect.

JTS22:
# Add support for returning alternative tokens

Fixes #298
So far, alternative tokens are only supported on `FlashCausalLM` based models.

## Before submitting
- [x] Was this discussed/approved via a Github issue or the discord / slack channel? Please add a link
      to it if that's the case: #298 

## Who can review?

### DIFF
--- a/server/lorax_server/models/flash_causal_lm.py
+++ b/server/lorax_server/models/flash_causal_lm.py
@@ -21,6 +21,7 @@ from lorax_server.models.cache_manager import (
 from lorax_server.models.types import (
     Batch,
     PrefillTokens,
+    AlternativeTokens,
     Generation,
     GeneratedText,
 )
@@ -844,6 +845,7 @@ class FlashCausalLM(Model):
     ) -> Tuple[List[Generation], Optional[FlashCausalLMBatch]]:
         prefill = batch.cu_seqlen_prefill is not None
         prefill_logprobs = batch.prefill_next_token_indices is not None
+        return_alternatives = any(req.parameters.return_k_alternatives > 0 for req in batch.requests)
 
         # Debugging for LoRAX
         # print("!!! adapter_indices", batch.adapter_indices)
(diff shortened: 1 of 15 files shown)


server/lorax_server/models/flash_causal_lm.py
```diff
@@ -1014,6 +1023,33 @@ def generate_token(
...
+                for alternative_token_id in batch_alternative_token_ids:
+                    all_input_ids.append(alternative_token_id)
+                    alternative_token_text, _, _ = self.decode_token(
+                        all_input_ids,
+                        prefix_offset,
+                        read_offset,
+                    )
+                    alternative_token_texts.append(alternative_token_text)
+                    all_input_ids.pop()
+                alternative_tokens = AlternativeTokens(
```
tgaddair, line 1045: It looks like we're overriding this at every iteration of the loop. Should `alternative_tokens` instead be a list of `AlternativeTokens`, one for each request in the batch?

tgaddair (commented):
Very clean and comprehensive PR, just had a question about how we handle multiple requests in the batch.

server/lorax_server/models/flash_causal_lm.py
JTS22: We are adding the `alternative_tokens` to the `Generation` that is created for each request, so I don't see a problem with overriding it. Basically I just copied from what is done with the `PrefillTokens`. The variables that go into `AlternativeTokens` are kind of badly named though, so I added another commit to fix that :)

tgaddair (approved):
LGTM! Thanks for this contribution :)

PR #299, opened 2024-03-04
Add support for LoRA adapters trained with Rank-Stabilized scaling

arnavgarg1:
In PEFT 0.9, support was added for adapters trained using a new flag called `use_rslora`.

When set to True, Rank-Stabilized LoRA sets the adapter scaling factor to `lora_alpha/math.sqrt(r)`, since it was proven to work better. Otherwise, it will use the original default value of `lora_alpha/r`.

In equation form:

- Normal LoRA layers: `W0X + (lora_alpha/r)(BAX)` where W0 is the base model, BA are the lora weight matrices and X is the input from the embedding layer/previous transformer layer.
- LoRA layers with RSLoRA enabled: W0X + (lora_alpha/sqrt(r))(BAX)

In particular, this is useful when using larger ranks since it prevents the gradient from collapsing as rank increases, which may result in higher ranks actually leading to better performance (not true by default today and in the original LoRA paper). Paper: https://arxiv.org/pdf/2312.03732.pdf.

image

### DIFF
--- a/server/lorax_server/models/model.py
+++ b/server/lorax_server/models/model.py
@@ -10,7 +10,12 @@ from transformers import PreTrainedTokenizerBase
 
 from lorax_server.models.types import Batch, GeneratedText
 from lorax_server.pb.generate_pb2 import AdapterParameters, AdapterSource, InfoResponse
-from lorax_server.utils.adapter import BASE_MODEL_ADAPTER_ID, load_and_merge_adapters
+from lorax_server.utils.adapter import (
+    BASE_MODEL_ADAPTER_ID,
+    get_scaling_factor,
+    load_and_merge_adapters,
+    uses_rslora,
+)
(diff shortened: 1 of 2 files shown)


server/lorax_server/models/model.py
```diff
@@ -217,25 +222,29 @@ def load_batched_adapter_weights(
...
-            
+
             lora_a, lora_a_name = module_map[weight_name]["lora_A"]
             lora_a = lora_a.to(base_device, self.dtype)
 
             lora_b, lora_b_name = module_map[weight_name]["lora_B"]
             lora_b = lora_b.to(base_device, self.dtype)
 
-            scale = adapter_config.lora_alpha / adapter_config.r
+            scale: float = get_scaling_factor(
```
tgaddair, line 243: can we remove `: float` here?

tgaddair (approved):

PR #303, opened 2024-03-04
Allow assigning dedicated memory reservation for adapters on GPU

tgaddair:
Related #283.

PR #306, opened 2024-03-06
Enforce adapters cannot be loaded past `--adapter-memory-fraction`

### LINKED ISSUE #53: Unexpected CUDA out of memory errors
### System Info

ghcr.io/predibase/lorax:latest

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [ ] My own modifications

### Reproduction

First, run

`docker run --gpus all --shm-size 1g -p 8000:80 -e "HUGGING_FACE_HUB_TOKEN=<token>" -v $volume:/data ghcr.io/predibase/lorax:latest --trust-remote-code --model-id meta-llama/Llama-2-7b-chat-hf`

Then, send the following requests:

```python
from concurrent.futures import ThreadPoolExecutor
import time
import requests

url = "http://127.0.0.1:8000/generate"
headers = {"Content-Type": "application/json"}

# Function to send a request
def send_request(payload):
    start_time = time.time()
    response = requests.post(url, headers=headers, json=payload)
    elapsed_time = time.time() - start_time
    return response.json(), elapsed_time

# Number of concurrent requests

adapter_list = [<list of 15 adapters>]
num_requests = len(adapter_list)

# Using ThreadPoolExecutor to send requests concurrently
with ThreadPoolExecutor(max_workers=num_requests) as executor:
    # Submit the requests
    futures = []
    for i in range(num_requests):
        payload = {
            "inputs": "Hello, my name is",
            "parameters": {"max_new_tokens": 100, "adapter_id": adapter_list[i]},
        }
        futures.append(executor.submit(send_request, payload))

    # Wait for all requests to complete
    results = [future.result() for future in futures]

print(results)
```

### Expected behavior

First of all, awesome work with lorax! When sending requests at the same time like this, I'm receiving Cuda Out Of Memory errors on the server. I thought that actually, the server would check beforehand how many tokens can be handled and consequently enqueue requests that cannot be served. Have you encountered this before?

tgaddair:
Fixes #53.

One of the main issues with our existing adapter loading and offloading strategy is that it relies on the user setting a fixed adapter limit via `--max-active-adapters` (default: `128`). However, this doesn't account for the fact that adapter sizes can vary by orders of magnitude based on the rank. As such, the server might do well with 128 rank 8 adapters but fall over with just a handful of rank 128 adapters.

In #303, we introduced a new parameter `--adapter-memory-fraction` that allows setting aside a dedicated pool of GPU memory to account for adapter overhead. This prevents the KV cache from expanding past the reservation set aside for adapters, reducing memory pressure. However, because the LoRAX scheduler still works off of the max active adapters, this means that users can still blow up the GPU memory by setting max active adapters higher than what can be accommodated by the adapter memory fraction.

This PR reconciles the scheduler with the adapter memory fraction. Now, the LoRAX scheduler will look at the size of each adapter after download to determine whether it can be loaded safely on the GPU. If not, then the adapter will wait until enough space is freed up the other active adapters becoming idle before the adapter can be moved to the GPU. This should ensure that no CUDA OOMs occur due to loading too many adapters.

Note that with this change, the max active adapters may no longer be needed, but we will keep it around for now to avoid backwards incompatible changes. However, the new default of 1024 should be sufficiently high that it won't be used in most cases.

### DIFF
--- a/router/src/queue.rs
+++ b/router/src/queue.rs
@@ -66,6 +66,9 @@ pub(crate) struct QueueState {
     /// Adapter status
     status: AdapterStatus,
 
+    /// Cost as a fraction of the adapter memory budget
+    cost: Option<f32>,
+
     /// Timestamp when the adapter was last activated
     activation_ts: Option<Instant>,
 
@@ -80,6 +83,7 @@ impl QueueState {
             entries: VecDeque::with_capacity(128),
             adapter,
             status,
+            cost: None,
             activation_ts: None,
             event,
         }
@@ -143,6 +147,14 @@ impl QueueState {
         &self.status
     }
 
+    pub(crate) fn set_cost(&mut self, cost: f32) {
+        self.cost = Some(cost);
+    }
+
+    pub(crate) fn cost(&self) -> Option<f32> {
+        self.cost
+    }
+
     pub(crate) fn set_activation_ts(&mut self, ts: Instant) {
         self.activation_ts = Some(ts);
     }
@@ -169,6 +181,9 @@ pub(crate) struct AdapterQueuesState {
     /// Number of adapters that can be active at a time
     max_active_adapters: usize,
 
+    /// Fraction of adapter memory budget remaining to allocate to new adapters
+    memory_budget_remaining: f32,
+
     /// Maximum time an adapter is allowed to be active before exchanging out
     max_active_time: Duration,
 
@@ -189,6 +204,7 @@ impl AdapterQueuesState {
             active_adapters,
             tracked_adapters,
             max_active_adapters: max_active_adapters,
+            memory_budget_remaining: 1.0,
             max_active_time: Duration::from_secs(adapter_cycle_time_s),
             next_id: 0,
         }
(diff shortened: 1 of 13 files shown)


server/lorax_server/models/model.py
```diff
@@ -81,6 +81,9 @@ def info(self) -> InfoResponse:
     @abstractmethod
     def batch_type(self) -> Type[B]:
         raise NotImplementedError
+    
+    def adapter_memory_size(self) -> int:
```
jeffreyftang, line 85: Would it be safer to raise `NotImplemented` instead?

tgaddair: We do want to return 0 here for other model types to ensure they still work (like Bloom), even though we won't be enforcing the adapter memory reservation.

router/src/queue.rs
```diff
@@ -388,20 +415,57 @@ impl AdapterQueuesState {
             }
         }
 
+        // Add back cost for all offload adapters
+        for adapter in offload_adapters.iter() {
+            let queue = self.queue_map.get(adapter).unwrap().clone();
+            let cost = queue.cost().unwrap();
```
jeffreyftang, line 421: Does this need a `none` check?

server/lorax_server/server.py
```diff
@@ -168,7 +176,26 @@ async def DownloadAdapter(self, request: generate_pb2.DownloadAdapterRequest, co
...
+                )
+        else:
+            # Assume 0.0 memory fraction if adapter memory size is not set
+            logger.info(f"Downloaded adapter {adapter_id} memory size: {adapter_bytes} bytes "
+                        f"(no reservation limit)")
+            adapter_memory_fraction = 0.0
+        
+        return generate_pb2.DownloadAdapterResponse(
+            downloaded=True,
+            memory_fraction=adapter_memory_fraction
```
jeffreyftang, line 197: Just curious - why convert to a fraction instead of passing the actual cost?

jeffreyftang (approved):

router/src/queue.rs
tgaddair: I had one originally, but we never add to the active set until the cost non-none (see check below). So if this condition is violated, it should be a programming error.

server/lorax_server/server.py
tgaddair: That way I don't have to plumb the reservation amount to the router, I just always work with 1 as the reservation amount for simplicity.

PR #309, opened 2024-03-06
Log generated tokens out to file when streaming

magdyksaleh:
This logs some request metadata to file when a streaming request ends. This allows us to better monitor usage and has some predibase uses. 

TODO: 
- [x] Make file location an ENV VAR 
- [ ]

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - 'main'
+      - 'stream-log-dump' # XXX
     tags:
       - 'v*'
 
@@ -61,11 +62,12 @@ jobs:
         with:
           images: |
             ghcr.io/predibase/lorax
+          # XXX
           tags: |
-            type=semver,pattern={{version}}
-            type=semver,pattern={{major}}.{{minor}}
-            type=sha,prefix=,suffix=,format=short
-            type=raw,value=latest
+            type=sha,prefix=magdy-test-,suffix=,format=short
+            
+            
+
       
       - name: Create a hash from tags
         env:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,7 +4,6 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - 'stream-log-dump' # XXX
       - 'main'
     tags:
       - 'v*'
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - 'main'
-      - 'stream-log-dump' # XXX
     tags:
       - 'v*'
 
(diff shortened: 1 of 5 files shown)


router/src/server.rs
```diff
@@ -733,6 +782,17 @@ async fn metrics(prom_handle: Extension<PrometheusHandle>) -> String {
     prom_handle.render()
 }
 
+async fn request_logger(mut rx: mpsc::Receiver<(i64, String, String)>) {
+    // log the function is getting called
+    let mut file = tokio::fs::File::create("/data/log.txt").await.unwrap();
```
noyoshi, line 787: Is there a way to ensure that this log file doesn't grow too large? E.g. maybe we partition the logs based on time, and can have a way to delete the old files?

router/src/server.rs
```diff
@@ -733,6 +784,34 @@ async fn metrics(prom_handle: Extension<PrometheusHandle>) -> String {
     prom_handle.render()
 }
 
+async fn request_logger(mut rx: mpsc::Receiver<(i64, String, String)>) {
+    let url = std::env::var("REQUEST_LOGGER_URL").ok();
+    if Some(&url) == None {
+        tracing::warn!("REQUEST_LOGGER_URL not set, request logging is disabled");
```
tgaddair, line 790: Should this be info? Seems like most users will be running in this state.

router/src/server.rs
```diff
@@ -927,15 +1006,17 @@ pub async fn run(
...
-    });
     DEFAULT_ADAPTER_SOURCE
         .set(adapter_source.clone())
         .unwrap_or_else(|_| {
             panic!("DEFAULT_ADAPTER_SOURCE was already set!");
         });
 
+    // Kick off thread here that writes to the log file
+    let (tx, rx) = mpsc::channel(32);
+    tokio::spawn(request_logger(rx));
```
tgaddair, line 1017: Should this be conditioned on `REQUEST_LOGGER_URL` being set?

magdyksaleh: Makes sense!

.github/workflows/build.yaml
```diff
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
   push:
     branches:
+      - 'stream-log-dump' # XXX
```
tgaddair, line 7: Let's remove this before landing.

router/src/server.rs
```diff
@@ -682,6 +731,8 @@ async fn generate_stream_with_callback(
                                         tracing::debug!(parent: &span, "Output: {}", output_text);
                                         tracing::info!(parent: &span, "Success");
 
+                                        request_logger_sender.send((generated_text.generated_tokens as i64, api_token.unwrap_or("".to_string()), info.model_id.clone())).await;
```
tgaddair, line 734: Any way we can make the `request_logger_sender` Optional so we don't do this when request logging is disabled? I'm not sure it will cause any issues, but it would be one less vector for things to go wrong (if, for example, we start accumulating a large queue of messages that never gets consumed, etc.).

router/src/server.rs
```diff
@@ -457,6 +512,16 @@ async fn generate(
         response.generated_text.generated_tokens as f64
     );
 
+    if std::env::var("REQUEST_LOGGER_URL").ok().is_some() {
```
tgaddair, line 515: We grab this env var in a lot of place using a raw string. Can we move this to a single place where we grab it once during init and then just reuse a variable like `request_logger_url.is_some()` from then on?

magdyksaleh: good point!

tgaddair (approved):
LGTM!

PR #320, opened 2024-03-11
Batch prefill tokens uses max input tokens as default

noyoshi:
`max_batch_prefill_tokens` is now optional, and it is defaulting to the value from `max_input_length`. 

Now, users can define custom `max_input_length` without having to also specify `max_batch_prefill_tokens`

### DIFF
--- a/launcher/src/main.rs
+++ b/launcher/src/main.rs
@@ -1183,6 +1183,13 @@ fn main() -> Result<(), LauncherError> {
 
     tracing::info!("{:?}", args);
 
+    // Set default values dervided from other args
+
+    // If the value of max_batch_prefill_tokens is not specified, default to max_input_length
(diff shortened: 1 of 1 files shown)


launcher/src/main.rs
```diff
@@ -1181,6 +1183,13 @@ fn main() -> Result<(), LauncherError> {
 
     tracing::info!("{:?}", args);
 
+    // Set default values dervided from other args
+
+    // If the value of max_batch_prefill_tokens is not specified, default to max_input_length
+    if args.max_batch_prefill_tokens.is_none() {
```
tgaddair, line 1189: Can we follow the same convention used by `max_batch_total_tokens`, which is also optional?

magdyksaleh:
@noyoshi how we looking here?

PR #324, opened 2024-03-13
Fix tensor parallelism with SGMV to use true rank of the LoRA after splitting

### LINKED ISSUE #308: SGMV not working for llama 70b
### System Info

```
lorax-launcher --env
2024-03-06T13:20:49.195306Z  INFO lorax_launcher: Runtime environment:
Target: x86_64-unknown-linux-gnu
Cargo version: 1.74.0
```

nvdiia-smi: 

```
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.104.05             Driver Version: 535.104.05   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA A100-SXM4-80GB          On  | 00000000:0F:00.0 Off |                    0 |
| N/A   53C    P0             271W / 400W |  79931MiB / 81920MiB |    100%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   1  NVIDIA A100-SXM4-80GB          On  | 00000000:15:00.0 Off |                    0 |
| N/A   53C    P0             377W / 400W |  80075MiB / 81920MiB |    100%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   2  NVIDIA A100-SXM4-80GB          On  | 00000000:50:00.0 Off |                    0 |
| N/A   52C    P0             382W / 400W |  77865MiB / 81920MiB |    100%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
|   3  NVIDIA A100-SXM4-80GB          On  | 00000000:53:00.0 Off |                    0 |
| N/A   56C    P0             434W / 400W |  79931MiB / 81920MiB |    100%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
```

model being used: llama-2-70b 

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [ ] My own modifications

### Reproduction

Just build the docker from main and try to infer lora adapter for 70b llama model with rank 8 for q.k.v.o matrices. 

I prepared a script to create folder with similar _adapter_ (just tensors of same shape). I attached [script](https://github.com/predibase/lorax/issues/241#issuecomment-1967433603) and details to previous issue. 

There is also an issue with very slow adapter loading: it takes more than 20seconds to load 30mb adapter, which is too slow (adapter is read from local ssd, no other tasks are running on the same node). 

It leads to significant increase in latency, meaning that (even with fast adapter loading) your latency for 1st token with 0.1 qps is more than 15seconds for 5 adapters, which is incredibly slow. 

### Expected behavior

SGMV kernels are utilised and lora inference works fast. 
Adapter weights are sharded in less than 200ms and not 20seconds

tgaddair:
Fixes #308.

Previous implementation assumed the rank set in the config was correct and we only need to scale by process group size. However, this may not always be the case depending on how the weights get split for tensor parallelism (column vs row parallel). As such, we should ignore the config and instead rely on the true rank dimension of the tensor independent of the process group size.

### DIFF
--- a/server/lorax_server/models/model.py
+++ b/server/lorax_server/models/model.py
@@ -263,11 +263,9 @@ class Model(ABC):
 
         if lora_a_list:
             # update rank if it was padded
-            print("PADDED RANK", lora_a_list[0].shape, lora_b_list[0].shape)
             padded_rank = lora_a_list[0].size(1)
             adapter_config.r = padded_rank
 
-        print("ADAPTER CONFIG", adapter_config.r)
         q_lora_merged = MergedLoraWeights(
             *self.shard_lora_weights(lora_a_list, lora_b_list, layer_type), adapter_config,
         )
--- a/server/lorax_server/models/model.py
+++ b/server/lorax_server/models/model.py
@@ -261,11 +261,13 @@ class Model(ABC):
         lora_a_list = [pad_rank(w, dim=1, world_size=self.world_size) for w in lora_a_list]
         lora_b_list = [pad_rank(w, dim=0, world_size=self.world_size) for w in lora_b_list]
 
-        if lora_b_list:
+        if lora_a_list:
             # update rank if it was padded
-            padded_rank = lora_b_list[0].size(0)
+            print("PADDED RANK", lora_a_list[0].shape, lora_b_list[0].shape)
+            padded_rank = lora_a_list[0].size(1)
             adapter_config.r = padded_rank
 
+        print("ADAPTER CONFIG", adapter_config.r)
         q_lora_merged = MergedLoraWeights(
             *self.shard_lora_weights(lora_a_list, lora_b_list, layer_type), adapter_config,
         )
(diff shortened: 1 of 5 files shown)


server/lorax_server/models/model.py
```diff
@@ -261,9 +261,9 @@ def load_batched_adapter_weights(
         lora_a_list = [pad_rank(w, dim=1, world_size=self.world_size) for w in lora_a_list]
         lora_b_list = [pad_rank(w, dim=0, world_size=self.world_size) for w in lora_b_list]
 
-        if lora_b_list:
+        if lora_a_list:
```
jeffreyftang, line 264: Curious what the significance of swapping from `lora_b_list` to `lora_a_list` here is.

jeffreyftang (approved):

server/lorax_server/models/model.py
tgaddair: It is the same, actually.

Issue #229, opened 2024-02-08
Load adapter from user-defined S3 service


codybum:
### Feature request

Provide more granular methods to configure the embedded S3 client.

### Motivation

Some users have their own S3 services and it would be useful to have a method to provide custom S3 client configurations.  Specifically, we would like to be able to load adapters from S3 buckets provided by our own S3 servers. 

### Your contribution

We could provide a PR to enable the minimal configuration needed, which is specifying the S3 service url_endpoint.  This is already partially done here: https://github.com/predibase/lorax/blob/5ca03154676ba5a30f466c010f37497c7f46bd26/server/lorax_server/utils/sources/s3.py#L65

At a minimum something like this should work:

S3_ENDPOINT_URL = os.environ.get("S3_ENDPOINT_URL", None)
    if S3_ENDPOINT_URL:
        s3 = boto3.resource('s3', 
            endpoint_url = f'{S3_ENDPOINT_URL}',
            config=config
           )
        return s3.Bucket(bucket_name)
    else:
        s3 = boto3.resource('s3', config=config)
        return s3.Bucket(bucket_name)

codybum:
Would it be helpful if I made a pull request, or does someone on the development team want to do it?

tgaddair:
Hey @codybum, apologies for the delay, we're catching up on a few things at the moment. If you have the bandwidth to put together a PR for this, that would be much appreciated. Thanks.

codybum:
Issue resolved with linked PR.

PR #325, opened 2024-03-13
Provide more granular methods to configure the embedded S3 client.

mitchklusty:
Referencing #229 for loading an adapter from user-defined S3 service.


tgaddair (approved):
Thanks for the PR! LGTM

PR #365, opened 2024-03-29
Update tagging logic and add flake8 linter

magdyksaleh:


.github/workflows/build.yaml
```diff
@@ -65,7 +65,7 @@ jobs:
             type=semver,pattern={{version}}
             type=semver,pattern={{major}}.{{minor}}
             type=sha,prefix=,suffix=,format=short
-            type=raw,value=latest
+            type=raw,value=main,enable=${{ github.ref == 'refs/heads/main' }}
```
magdyksaleh, line 68: we get the `latest` tag by default for semver type events as per https://github.com/docker/metadata-action?tab=readme-ov-file#latest-tag

We just add the main tag now when the we build due to a push to main

PR #377, opened 2024-04-03
fix case where we have empty lora weights when using medusa

noyoshi:
Whenever we try to use medua weights with more than 1 new token, we get 
```Server error: 'list' object has no attribute 'rank_data'```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - 'main'
+      - 'fix-empty-lora-weights'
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: .github/workflows/build.yaml [hunk 1]
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - 'main'
+      - 'fix-empty-lora-weights'
```
tgaddair, line 8: Please remove before landing.

tgaddair:
Resolved by #381.

PR #403, opened 2024-04-09
enh: Update Predibase integration to support v2 API

jeffreyftang:
Predibase is releasing a new API for working with and retrieving adapters. This PR updates Lorax's Predibase integration to prefer the new API, falling back to the legacy endpoints to maintain backwards compatibility during the transition period.

### DIFF
--- a/server/lorax_server/utils/sources/__init__.py
+++ b/server/lorax_server/utils/sources/__init__.py
@@ -50,7 +50,7 @@ def map_pbase_model_id_to_s3(model_id: str, api_token: str) -> str:
             # Try to retrieve data using the new endpoint.
             resp = requests.get(url, headers=headers)
             resp.raise_for_status()
-        except:  # ruff: noqa E722
+        except requests.RequestException:
             # Not found in new path, fall back to legacy endpoint.
             resp = requests.get(legacy_url, headers=headers)
             resp.raise_for_status()
--- a/server/lorax_server/utils/sources/__init__.py
+++ b/server/lorax_server/utils/sources/__init__.py
@@ -19,8 +19,9 @@ S3 = "s3"
 LOCAL = "local"
 PBASE = "pbase"
 
-PREDIBASE_MODEL_URL_ENDPOINT = "/v1/models/version/name/{}"
-PREDIBASE_MODEL_VERSION_URL_ENDPOINT = "/v1/models/version/name/{}?version={}"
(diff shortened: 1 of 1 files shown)


server/lorax_server/utils/sources/__init__.py
```diff
@@ -30,19 +31,42 @@ def map_pbase_model_id_to_s3(model_id: str, api_token: str) -> str:
...
+            resp.raise_for_status()
+            uuid, best_run_id = resp.json()["uuid"], resp.json()["bestRunID"]
+            return f"{uuid}/{best_run_id}/artifacts/model/model_weights/"
+
+        path = resp.json().get("adapterPath", None)
+        if path is None:
+            raise RuntimeError(f"Adapter {model_id} is not yet available")
+        return path
+    else:
+        # Use legacy path only since new endpoint requires both name and version number.
```
tgaddair, line 65: Looks like this code is identical to 54:58. Can we pull it into a function?

```
def fetch_legacy_url(legacy_url):
    resp = requests.get(legacy_url, headers=headers)
    resp.raise_for_status()
    uuid, best_run_id = resp.json()["uuid"], resp.json()["bestRunID"]
    return f"{uuid}/{best_run_id}/artifacts/model/model_weights/"
```

server/lorax_server/utils/sources/__init__.py
```diff
@@ -30,19 +31,42 @@ def map_pbase_model_id_to_s3(model_id: str, api_token: str) -> str:
...
+        legacy_url = PREDIBASE_GATEWAY_ENDPOINT + LEGACY_PREDIBASE_MODEL_VERSION_URL_ENDPOINT.format(name, version)
     else:
         raise ValueError(f"Invalid model id {model_id}")
-    resp = requests.get(url, headers=headers)
-    resp.raise_for_status()
-    uuid, best_run_id = resp.json()["uuid"], resp.json()["bestRunID"]
-    return f"{uuid}/{best_run_id}/artifacts/model/model_weights/"
+
+    if url is not None:
+        try:
```
martindavis, line 49: Just trying to think about this in terms of customer impact: we really only expect staging to use this for the next few days until we migrate all of LLM Models to the Adapters table and a very brief window in Prod while we migrate the data right?

martindavis (approved):

server/lorax_server/utils/sources/__init__.py
jeffreyftang: Done.

tgaddair (approved):
LGTM!

PR #405, opened 2024-04-10
Bump pydantic version to >2, <3

claudioMontanari:
# What does this PR do?

Fixes #177 

I am running `poetry lock --no-update` to patch the `poetry.lock` file. Please, advise if you have a preferred alternative! 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Was this discussed/approved via a Github issue or the discord / slack channel? Please add a link
      to it if that's the case. https://github.com/predibase/lorax/issues/177
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.

PR #406, opened 2024-04-10
int: Upgrade lorax-client to pydantic v2

jeffreyftang:
This PR bumps the required pydantic version, but imports from `pydantic.v1`, so functionality should remain the same.


jeffreyftang:
Closing in favor of #405

PR #419, opened 2024-04-16
lorax launcher now has --default-adapter-source

noyoshi:
The `--adapter-source` is the adapter source used for loading static adapters. If `--default-adapter-source` is not given, it is also used for loading in the dynamic adapters. 

If you want a different default for the dynamic adapters, then use `--default-adapter-source` instead :)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,154 +1,154 @@
 name: Build and push docker image to ghcr
 
 on:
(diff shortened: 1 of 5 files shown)


tgaddair (commented):
Can we flip this around:

`--adapter-source` applies to the static adapter and `--default-adapter-source` is the default used per request?

tgaddair (approved):

PR #428, opened 2024-04-19
use default HF HUB token when checking for base model info

noyoshi:
- This is important now that the models are gated, we need to use the token which started the model, since that _should_ have permissions to read the base model repo!

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,154 +1,154 @@
 name: Build and push docker image to ghcr
 
 on:
-    workflow_dispatch:
-    push:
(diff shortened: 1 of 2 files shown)


magdyksaleh (approved):

tgaddair (approved):

bi1101:
Hi, does this PR affect the Predibase API? I've been getting error with inferencing finetuned models through the OAI API since roughly 12 hours ago
```
{
    "error": "Request failed during generation: Server error: Unable to locate credentials",
    "error_type": "generation"
}
```

PR #431, opened 2024-04-21
fix: checking the base_model_name_or_path of adapter_config and early return if null

### LINKED ISSUE #430: 10s latency of lora inference caused by None base_model_name_or_path in adapter_config
### System Info

lorax: main or v0.9.0

### Information

- [ ] Docker
- [ ] The CLI directly

### Tasks

- [ ] My own modifications

### Reproduction

1. launch the lorax
2. prepare one lora weights, with null `base_model_name_or_path` in adapter_config
3. inference with this adapter, but there is 10s delay to get first token

### Expected behavior

Expect to get first token within 1 seconds instead of 10 seconds.

thincal:
# What does this PR do?

Fixes #430 

## Before submitting
      to it if that's the case.

## Who can review?

@tgaddair

### DIFF
--- a/server/lorax_server/utils/adapter.py
+++ b/server/lorax_server/utils/adapter.py
@@ -86,6 +86,10 @@ def _load_and_merge(
 
 def check_architectures(model_id: str, adapter_id: str, adapter_config: "AdapterConfig"):
     try:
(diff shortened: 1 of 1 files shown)


tgaddair (approved):
Thanks for the fix!

PR #445, opened 2024-04-29
Added phi3

tgaddair:
```
    -X POST \
    -d '{
        "inputs": "<|user|>\nHow to explain Internet for a medieval knight?<|end|>\n<|assistant|>"
    }' \
```

Note that because the model is very small, compilation (`--compile`) had a noticeable affect on throughput (81 tok/s to 95 tok/s on A100 with default settings).

### DIFF
--- a/server/lorax_server/models/flash_phi3.py
+++ b/server/lorax_server/models/flash_phi3.py
@@ -3,7 +3,7 @@ from typing import Dict, List, Optional, Tuple
 import torch
 import torch.distributed
 from opentelemetry import trace
-from transformers import AutoTokenizer
+from transformers import AutoTokenizer, GenerationConfig
 
 from lorax_server.models import FlashCausalLM
 from lorax_server.models.custom_modeling.flash_phi3_modeling import (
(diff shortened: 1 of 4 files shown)


server/lorax_server/models/flash_phi3.py
```diff
@@ -58,6 +58,19 @@ def __init__(
...
+            # if it is a list or set. We need to do this by adding a new property as the tokenizer
+            # does not officially support multiple eos_token_ids.
+            generation_config = GenerationConfig.from_pretrained(
+                model_id, revision=revision, trust_remote_code=trust_remote_code
+            )
+
+            if isinstance(generation_config.eos_token_id, (list, set)):
+                tokenizer.eos_token_ids = set(generation_config.eos_token_id)
+        except Exception:
+            pass
```
arnavgarg1: Super interesting that the tokenizer only has 1 EOS token ID but the generation config actually maps to a list of token IDs for eos_token_id. I assume this is to support Llama-3 right?

tgaddair: For whatever reason both Llama-3 and Phi-3 do this.

server/lorax_server/utils/tokens.py
```diff
@@ -188,8 +192,9 @@ def from_pb(
         tokenizer: PreTrainedTokenizerBase,
     ) -> "StoppingCriteria":
         stop_sequence_criterias = [StopSequenceCriteria(sequence) for sequence in pb.stop_sequences]
+        eos_token_id = getattr(tokenizer, "_eos_token_ids", tokenizer.eos_token_id)
```
arnavgarg1, line 195: this should be `getattr(tokenizer, "eos_token_ids", tokenizer.eos_token_id)` right?

server/lorax_server/models/flash_phi3.py
arnavgarg1: Oh very strange that they have multiple stop tokens, but good to know!

server/lorax_server/utils/tokens.py
tgaddair: I was going back and forth but let's go with no _

arnavgarg1: I see, because we're setting a custom property on the tokenizer? both make sense! Maybe _ might be better actually

arnavgarg1 (approved):

PR #452, opened 2024-05-01
Updated lorax-launcher docs to include HF Hub API Token

### LINKED ISSUE #451: Add HF authentication instructions to lorax-launcher docs
On the [LoRAX launcher docs page](https://loraexchange.ai/reference/launcher/), a `MODEL_ID` attribute is accepted. However, if the model on Huggingface requires authentication, the current docs provide no solution for providing a Huggingface API token to authenticate this LoRAX launcher request.

Instructions on how to provide a HF token can be found [here](https://huggingface.co/docs/text-generation-inference/en/basic_tutorials/gated_model_access). Perhaps adding something like this on the LoRAX launcher page would be helpful.

kunal-mansukhani:
# What does this PR do?

Fixes #451 Add HF authentication instructions to lorax-launcher docs

## Before submitting
- [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).

## Who can review?

@tgaddair

### DIFF
--- a/docs/reference/launcher.md
+++ b/docs/reference/launcher.md
@@ -12,6 +12,11 @@ Options:
           [env: MODEL_ID=]
           [default: mistralai/Mistral-7B-Instruct-v0.1]
 
+      --env <HUGGING_FACE_HUB_TOKEN>
+          Your Hugging Face API token to authenticate access to private or gated models on the Hugging Face Hub. This token can be set as an environment variable or passed directly through the command line
+
+        [env: HUGGING_FACE_HUB_TOKEN=]
+
       --adapter-id <ADAPTER_ID>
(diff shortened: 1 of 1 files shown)


docs/reference/launcher.md
```diff
@@ -12,6 +12,11 @@ Options:
           [env: MODEL_ID=]
           [default: mistralai/Mistral-7B-Instruct-v0.1]
 
+      --env <HUGGING_FACE_HUB_TOKEN>
```
JamsheedMistri, line 15: Should be `--env HUGGING_FACE_HUB_TOKEN=<HUGGING_FACE_HUB_TOKEN>`, according to the rest of the documentation.

JamsheedMistri (changes requested):

docs/reference/launcher.md
tgaddair: Correct. This is also something that should be set in the container, not on the lorax launcher itself.

I can add some quick docs to reflect this.

PR #468, opened 2024-05-15
Reuse session per class instance

gyanesh-mishra:
# What does this PR do?

The PR introduces a "session" object to reuse when making API calls to the LLM server. Instead of instantiating a new requests instance on each HTTP call, reusing the same session object allows you to reuse socket connections and improve performance.

# How was this PR tested

I ran netstat to monitor the number of open sockets/connections the library makes before and after this change.
Before this change ~40 open socket sessions were created on a normal LLM call which would expire about every 30 seconds, after this change it makes 1 socket session that gets renewed/refreshed every time there is call to the API.

FIXES: INFRA-2744

### DIFF
--- a/clients/python/lorax/client.py
+++ b/clients/python/lorax/client.py
@@ -42,6 +42,7 @@ class Client:
         headers: Optional[Dict[str, str]] = None,
         cookies: Optional[Dict[str, str]] = None,
         timeout: int = 60,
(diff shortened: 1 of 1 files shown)


tgaddair (commented):
@gyanesh-mishra can you make this change for the AsyncClient as well?

noyoshi (approved):

PR #475, opened 2024-05-18
Fix for the LM_HEAD issue

### LINKED ISSUE #163: CUDA error when for sgmv_lora_b for LM_head with many concurrent requests
It seems that if the server is flooded with requests for a new adapter it needs to download, a race condition can arise leading to CUDA errors. Needs more investigation.

ajtejankar:
# (WIP) Fix for the LM_HEAD issue

**Root Cause**. The error is caused by incorrect segments passed to the `lora_b_sgmv` kernel during the prefill stage. This happens because we do not want to forward all the tokens in the prompt through `lm_head` and associated adapters. The goal is to save compute and memory by not having to forward the tokens that are not involved in generating the next token. Only the last token is needed for this purpose. However, doing this changes the shape of the net batch size (batch size times number of tokens) that the lora kernel sees, but the segment start and end tensors are not changed. These tensors are used to slice/segment the batch in the kernel. Hence, keeping them unchanged is incorrect. Additionally, the resulting error from this problem not reported correctly since the kernel has a catch-call condition that reports a generic `kernel not found for the dtype` message. My guess is that it's an out-of-bounds memory access, but confirming this would require changing the kernel. This is out of the scope of this investigation.

**Description of the Fix**. The fix simply goes over the adapters for `lm_head` and adjusts the segment start and end tensors to correctly point to the segments in the batch. For now, this is done only during the prefill stage and when a LoRA adapter for the `lm_head` is present.

Fixes #163 

## Before submitting
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/lorax_server/adapters/lora.py
+++ b/server/lorax_server/adapters/lora.py
@@ -162,6 +162,7 @@ class LoraWeights(AdapterWeights):
 @dataclass
 class RankSegments:
     rank: int
+    indices: List[int]
     tmp_shrink: torch.Tensor
     tmp_expand: torch.Tensor
     lora_a_ptr: torch.Tensor
@@ -231,6 +232,7 @@ class BatchLoraWeights(BatchAdapterWeights):
 
             rank_data[rank] = RankSegments(
                 rank=rank,
+                indices=indices,
                 tmp_shrink=tmp_shrink,
                 tmp_expand=tmp_expand,
                 lora_a_ptr=lora_a_ptr_indices,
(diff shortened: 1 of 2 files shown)


tgaddair (changes requested):
This was a good catch! In terms of making this more general, I think we can safely move this into the [BatchLoraWeights.load](https://github.com/predibase/lorax/blob/main/server/lorax_server/adapters/lora.py#L246) function which stores off unique `segment_indices` per layer.

One thing we would just need to change is to (1) plumb through the `batch.prefill_head_indices` [here](https://github.com/predibase/lorax/blob/main/server/lorax_server/models/flash_causal_lm.py#L955) and (2) plumb through the layer name (which is `k`) [here](https://github.com/predibase/lorax/blob/main/server/lorax_server/adapters/weights.py#L104).

Also, looks like there are a couple merge conflicts that need to be cleaned up.

ajtejankar:
Sounds good. Will make those changes and do a basic testing round with a few different model+LoRA combinations.

tgaddair (approved):
LGTM! Thanks for the fix!

PR #476, opened 2024-05-19
fix: load tokenizer/config with trust_remote_code

### LINKED ISSUE #466: AutoTokenzier.from_pretrains needs setting with `trust_remote_code` inside `load_module_map`
### System Info

lorax main

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [ ] My own modifications

### Reproduction

1. using a tokenizer class not registered in the Transformers
2. place the tokenizer config file and related files inside the adapter checkpoint folder
3. inference with adapter info pointed to this adapter folder
4. the tokenizer loading will go into the resolve_trust_remote_code logic inside the Transformers, where it will require user input to confirm if `trust_remote_code` is not True.

### Expected behavior

loading the tokenizer directly if it is existed inside adapter.

thincal:
# What does this PR do?

loading the tokenizer with remote code and unregistered class, will require user interactive input for confirming yes/no. That will break the normal processing.

Fixes #466

## Before submitting
      to it if that's the case.

## Who can review?


server/lorax_server/utils/adapter.py
```diff
@@ -90,8 +90,8 @@ def check_architectures(model_id: str, adapter_id: str, adapter_config: "Adapter
             # Avoid execuation latency caused by the network connection retrying for AutoConfig.from_pretrained(None)
             return
 
-        expected_config = AutoConfig.from_pretrained(model_id)
-        model_config = AutoConfig.from_pretrained(adapter_config.base_model_name_or_path)
+        expected_config = AutoConfig.from_pretrained(model_id, trust_remote_code=True)
```
tgaddair, line 93: Hey @thincal, the only concern I have is setting this to `True` by default exposes users to potential security vulnerabilities. We have a flag that users can provide for `--trust-remote-code` already. Can you plumb that through to this location so we don't need to default to setting it to True?

tgaddair (changes requested):
Thanks for the PR! Can we use the existing `--trust-remote-code` flag for this?

server/lorax_server/utils/adapter.py
thincal: it's reasonable, fixed.

tgaddair (approved):
LGTM! Thanks for the fix.

PR #495, opened 2024-06-03
Add retries on common session errors for the client

gyanesh-mishra:
# What does this PR do?

- Adds a retry adapter to catch the most common HTTP status codes and retry with exponential backoff
- Adds an internal POST method that refreshes session when remote server closes a long running session

### DIFF
--- a/clients/python/lorax/client.py
+++ b/clients/python/lorax/client.py
@@ -46,6 +46,7 @@ class Client:
         cookies: Optional[Dict[str, str]] = None,
         timeout: int = 60,
         session: Optional[requests.Session] = None,
+        max_session_retries: int = 2,
     ):
         """
         Args:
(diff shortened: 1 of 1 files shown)


clients/python/lorax/client.py
```diff
@@ -64,7 +65,72 @@ def __init__(
...
+        
+        Returns: 
+            requests.Response: HTTP response object
+        """
+        # Instantiate session if currently None
+        if not self.session:
+            self._create_session()
+
+        # Retry upto 2 times with a fresh session for Connection Errors
+        total_retries = 2
```
tgaddair, line 106: Can you make this configurable in the client initializer?

tgaddair (approved):

clients/python/lorax/client.py
gyanesh-mishra: Done ✅

PR #513, opened 2024-06-12
fix batching bug 

magdyksaleh:


tgaddair (approved):
Great catch! Definitely looks correct given what we were doing before:

https://github.com/predibase/lorax/commit/e8f3d33ff100da93069901f638c8173fb72baafe#diff-b1fa6727513f9184386a8371f182d51c9ccd6d131d61be6218170a6e2444146aL417-L420

PR #516, opened 2024-06-14
try out an integration test workflow

noyoshi:


.github/workflows/integration_tests.yaml
```diff
@@ -0,0 +1,41 @@
...
+      # Install the client 
+      - run: pip install -e clients/python
+      - name: create the pod and then delete the pod
+        run: |
+          bash tests/create-pod.sh ${{ inputs.lorax_tag }}
+      
+      - name: Run tests 
+        run: |
+          POD_ID=`cat pod_name.txt`
+          python3 tests/test.py $POD_ID
```
tgaddair, line 36: Where is this coming from?

tgaddair (approved):

.github/workflows/integration_tests.yaml
noyoshi: I might make it more obvious, its coming from `create-pod.sh`

I now see that I didn't commit that folder...

PR #519, opened 2024-06-18
try s3 crt

noyoshi:
attempts to use S3 crt to download weights to make downloads faster


noyoshi:
[313cb94](https://github.com/predibase/lorax/commit/313cb94d3cc7947afab1d8ebf7366b8188848174) works but I am not sure if it is using CRT :3

PR #530, opened 2024-06-30
Added Gemma2

tgaddair:
Model: https://huggingface.co/google/gemma-2-9b-it

Prompt:

```
    -X POST \
    -d '{
        "inputs": "<bos><start_of_turn>user\nWrite a hello world program<end_of_turn>\n<start_of_turn>model\n",
        "parameters": {
            "max_new_tokens": 64
        }
    }' \
```

### DIFF
--- a/server/lorax_server/models/flash_gemma2.py
+++ b/server/lorax_server/models/flash_gemma2.py
@@ -121,7 +121,7 @@ class FlashGemma2(FlashCausalLM):
         return [Q_PROJ, V_PROJ]
 
     def get_num_layers_for_type(self, layer_type: str) -> int:
-        return len(self.model.model.layers)
+        return 1 if layer_type == LM_HEAD else len(self.model.model.layers)
 
     def is_row_parallel(self, layer_type: str) -> bool:
         return layer_type in ROW_PARALLEL
--- a/server/lorax_server/models/flash_gemma2.py
+++ b/server/lorax_server/models/flash_gemma2.py
@@ -1,0 +1,128 @@
+from typing import Dict, List, Optional, Tuple
+
+import torch
+import torch.distributed
+from opentelemetry import trace
+from transformers import AutoTokenizer, PretrainedConfig
+
+from lorax_server.models import FlashCausalLM
+from lorax_server.models.custom_modeling.flash_gemma2_modeling import FlashGemma2ForCausalLM
+from lorax_server.utils import (
+    Weights,
+    initialize_torch_distributed,
+    weight_files,
+)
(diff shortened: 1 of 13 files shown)


server/lorax_server/utils/paged_attention.py
```diff
@@ -14,11 +14,14 @@
...
-    fp8_supported = torch.cuda.get_device_capability()[0] >= 9 #or (torch.cuda.get_device_capability()[0] == 8 and torch.cuda.get_device_capability()[1] >= 9)
-else:
-    fp8_supported = False
+# TODO(travis): fix for CUDA 8.9 (Lovelace) and 9.0 (Hopper)
+# if torch.cuda.is_available():
+#     fp8_supported = (
+#         torch.cuda.get_device_capability()[0] >= 9
+#     )  # or (torch.cuda.get_device_capability()[0] == 8 and torch.cuda.get_device_capability()[1] >= 9)
+# else:
+fp8_supported = False
```
arnavgarg1, line 23: Can we also add this same comment to `server/lorax_server/models/flash_causal_lm.py` or just refactor to use that?

server/lorax_server/models/flash_gemma2.py
```diff
@@ -0,0 +1,127 @@
...
+    @property
+    def adapter_layers(self) -> List[str]:
+        return ADAPTER_LAYERS
+
+    @property
+    def default_traced_adapter_layers(self) -> List[str]:
+        return [Q_PROJ, V_PROJ]
+
+    def get_num_layers_for_type(self, layer_type: str) -> int:
+        return len(self.model.model.layers)
```
arnavgarg1, line 124: Can also just return this from the config if we want to make the config stateful, otherwise this is fine too! https://huggingface.co/google/gemma-2-9b/blob/main/config.json#L21

arnavgarg1: Also should this be:

```
return 1 if layer_type == LM_HEAD else len(self.model.model.layers)
```

so we utilize the layer_type arg?

server/lorax_server/utils/paged_attention.py
tgaddair: Not sure I follow, where would you want it added in `flash_casual_lm.py`?

server/lorax_server/models/flash_gemma2.py
tgaddair: Yes, good catch!

server/lorax_server/utils/paged_attention.py
tgaddair: Ah, okay I see. That seems to be unused code. Removed.

arnavgarg1 (approved):

PR #534, opened 2024-07-01
Infer dtype from model config when not explicitly specified

arnavgarg1:
Instead of defaulting to float16, we should default to the dtype explicitly specified in the config of the associated model since some models like Gemma-2 and even mistral are sensitive to it and benefit from the dtype they were trained in (bfloat16 instead of float16)

### DIFF
--- a/server/lorax_server/models/__init__.py
+++ b/server/lorax_server/models/__init__.py
@@ -81,7 +81,7 @@ def get_model(
     elif dtype == "bfloat16":
         if not is_bf16_supported():
             if is_dtype_provided:
-                raise RuntimeError("bfloat16 is not supported on this device, set dtype to float16.")
+                raise RuntimeError("bfloat16 is not supported on this device, set --dtype float16.")
             logger.warning("bfloat16 is not supported on this device, falling back to float16")
             dtype = torch.float16
         else:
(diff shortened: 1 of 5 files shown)


server/lorax_server/models/__init__.py
```diff
@@ -72,15 +73,21 @@ def get_model(
...
     elif dtype == "bfloat16":
-        dtype = torch.bfloat16
+        if is_bf16_supported():
+            dtype = torch.bfloat16
+        else:
+            logger.warning("bfloat16 is not supported on this device, falling back to float16")
+            dtype = torch.float16
     else:
-        raise RuntimeError(f"Unknown dtype {dtype}")
+        try:
```
tgaddair, line 87: If the user explicitly specified bf16 but it failed, that seems like it should be an error

tgaddair (commented):
Did you also check that this is set correctly in the Rust code?

server/lorax_server/models/__init__.py
arnavgarg1: Agreed

arnavgarg1:
> Did you also check that this is set correctly in the Rust code?

@tgaddair are you able to elaborate? As far as I can tell, the rust server seems to understand the notion of both float16 and bfloat16 dtypes, and I can launch both the server + router when the dtype is bfloat16 (inferred or explicitly specified for the server at launch). 

Let me know if I am missing some information on how things work!

Infernaught (approved):

ajtejankar (approved):

PR #559, opened 2024-07-27
Support FP8 for Mistral

ajtejankar:
Supports running the Mistral model with fused operations. The accuracy is the same as the unfused implementation, but speed is slow as this implementation uses cutlass kernels which are known to be slow on Ada GPUs. Need to test on H100s to fully test the implementation.


tgaddair:
@ajtejankar can you fix formatting?

server/lorax_server/utils/weights.py
```diff
@@ -116,8 +116,33 @@ def get_multi_weights_col(self, prefixes: List[Union[str, Tuple]], quantize: str
             bits, groupsize = self._get_bits_and_groupsize()
             weight = (qweight, qzeros, scales, g_idx, bits, groupsize, False)
         else:
-            w = self.get_sharded_list("weight", prefixes, dim=0)
-            weight = torch.cat(w, dim=dim)
+            weight_list = self.get_sharded_list("weight", prefixes, dim=0)
```
tgaddair, line 119: Can we unit-test this logic?

tgaddair:
@ajtejankar looks like there are still ruff errors. Can you take a look?

ajtejankar:
Yes, working on it. I don't know why vllm import is failing during the tests.

ajtejankar:
@tgaddair Do you know what's happening with the tests? I have no clue why vllm import would fail.

tgaddair:
@ajtejankar the main issue there is the vllm isn't installed for unit tests (as it's built as part of the GPU kernel compile process). So I moved the `is_fp8_quantized` helper into a separate file to the vllm import is off the critical path.

tgaddair (approved):

ajtejankar:
Thanks! I am working on some unit tests. Once they are done, I will merge the PR.

PR #560, opened 2024-07-28
Fixes

magdyksaleh:
Fix ner on non-homogeneous batches

### DIFF
--- a/router/client/src/sharded_client.rs
+++ b/router/client/src/sharded_client.rs
@@ -157,6 +157,7 @@ impl ShardedClient {
     /// Embed the given batch
     #[instrument(skip(self))]
     pub async fn embed(&mut self, batch: Batch) -> Result<Vec<Embedding>> {
+        tracing::info!("Embedding batch");
         let futures: Vec<_> = self
             .clients
             .iter_mut()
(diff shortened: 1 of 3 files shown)


router/client/src/sharded_client.rs
```diff
@@ -157,6 +157,7 @@ impl ShardedClient {
     /// Embed the given batch
     #[instrument(skip(self))]
     pub async fn embed(&mut self, batch: Batch) -> Result<Vec<Embedding>> {
+        tracing::info!("Embedding batch");
```
tgaddair, line 160: Can you make this debug or trace level?

server/lorax_server/models/types.py
```diff
@@ -163,6 +163,7 @@ def from_pb(
             return_token_type_ids=True,
             truncation=True,
             max_length=max_truncation,
+            padding=True,
```
tgaddair, line 166: Hmm, is padding necessary? My understanding is flash attention should not require padding.

tgaddair: Also, we should be doing tokenization on the Rust side now, so you should have access to the tokenized data from the request payload (see `flash_causal_lm.py` implementation).

magdyksaleh: you need to pad the input tokens since we are feeding them into an encoder layer after embedding that does them by batch and so we need (batch, seq_len, hidden) dims

PR #561, opened 2024-07-28
Move NER output formatting to server 

magdyksaleh:


router/client/src/sharded_client.rs
```diff
@@ -268,3 +307,68 @@ fn merge_generations(
...
+        re.find_iter(&tokenizer.decode(input_ids, true).unwrap())
+            .map(|m| m.as_str().to_string())
+            .collect()
+    };
+
+    let mut ner_results = Vec::new();
+    let mut current_entity: Option<Entity> = None;
+    for (i, ((token, token_class), score)) in tokens
+        .iter()
+        .zip(predicted_token_class.iter())
```
jeffreyftang, line 332: Btw, looks like there's a nice macro for zipping multiple iterators: https://stackoverflow.com/a/29669741

magdyksaleh: oooh good find

router/client/src/sharded_client.rs
```diff
@@ -174,8 +184,37 @@ impl ShardedClient {
             .iter_mut()
             .map(|client| Box::pin(client.classify(batch.clone())))
             .collect();
-        let results: Result<Vec<Vec<EntityList>>> = join_all(futures).await.into_iter().collect();
-        Ok(results?.into_iter().flatten().collect())
+        let results: Result<Vec<Vec<ClassifyPredictionList>>> =
+            join_all(futures).await.into_iter().collect();
+
+        let flat_results: Vec<ClassifyPredictionList> = results?.into_iter().flatten().collect();
+        let threaded_futures: Vec<_> = flat_results
```
tgaddair, line 191: I would be very curious to time this block of code to see how long it takes. My concern with this approach is that even though we're parallelizing the `format_ner_outputs` operation, we're doing so in a way that is blocking the batching thread, so during this period of time, the GPU is entirely idle.

Instead, we should move this logic into the request thread here:

https://github.com/predibase/lorax/blob/main/router/src/infer.rs#L659-L661

In that way, we do not block the batching thread so we can move on to the next batch and keep the GPU busy.

Does that make sense? Happy to quickly sync on this if needed.

magdyksaleh: addressed

router/src/server.rs
```diff
@@ -1079,6 +1079,7 @@ pub async fn run(
     struct ApiDoc;
 
     let cloned_tokenizer = tokenizer.clone().map(|t| Arc::new(Mutex::new(t)));
+    let arc_tokenizer = tokenizer.clone().map(Arc::new);
```
tgaddair, line 1082: Is it not sufficient to use the `cloned_tokenizer` here?

tgaddair (approved):
Nice! Do you have benchmarking results comparing this change with the previous version (decoding in the batching thread)?

magdyksaleh:
> Nice! Do you have benchmarking results comparing this change with the previous version (decoding in the batching thread)?

Yep: slightly better performance when moving the output formatting out of the batching task. https://predibase.slack.com/archives/C03MMJ8J5GF/p1722357582611049?thread_ts=1722331652.853819&cid=C03MMJ8J5GF

PR #562, opened 2024-07-31
(WIP) Support FP8 for LLaMa

ajtejankar:
1. Add support for FP8 weights in llama
2. Fix the bug where input scales are not converted to fp32

### DIFF
--- a/server/lorax_server/layers/fp8.py
+++ b/server/lorax_server/layers/fp8.py
@@ -36,7 +36,7 @@ class Fp8Linear(torch.nn.Module):
         self.qweight = weight.t()
         self.weight_scale = weight_scale.view(1, -1).contiguous().float()
         self.qbias = bias if bias is not None else None
-        self.input_scale = input_scale.float() if input_scale else input_scale
+        self.input_scale = input_scale.float() if input_scale is not None else None
 
     def forward(self, input: torch.Tensor) -> torch.Tensor:
         return apply_fp8_linear(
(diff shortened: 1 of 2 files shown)


server/lorax_server/models/custom_modeling/flash_llama_modeling.py
```diff
@@ -195,7 +195,11 @@ def _load_gqa(config, prefix: str, weights):
         dim=0,
     )
 
-    if config.quantize not in ["gptq", "awq"]:
+    input_scale, weight_scale = None, None
+    if type(weight) is tuple:
```
tgaddair, line 199: Style: please use isinstance instead

server/lorax_server/layers/fp8.py
```diff
@@ -34,9 +34,9 @@ def __init__(
         super().__init__()
         self.dtype = weight.dtype
         self.qweight = weight.t()
-        self.weight_scale = weight_scale.view(1, -1).contiguous()
+        self.weight_scale = weight_scale.view(1, -1).contiguous().float()
         self.qbias = bias if bias is not None else None
-        self.input_scale = input_scale
+        self.input_scale = input_scale.float() if input_scale else input_scale
```
tgaddair, line 39: Is this a check for None? If so, can we be more explicit in that check:

x.float() if x is not None else None

tgaddair (approved):

tgaddair (commented):
Few nits

tgaddair (approved):

PR #579, opened 2024-08-20
Introduce the "adapter_version" parameter for convenience in the "generate()" methods.

alexsherstinsky:
# What does this PR do?

For user and notebook use-cases, one often works with fine-tuned adapters hosted by Predibase.  The `adapter_id` format for such adapters is: repository/version where repository (string) and version (int) are both required (with the "/" as the separator between).  However, this causes users to constantly have to put in an "if then": without the adapter, `adapter_id` can be omitted; with the adapter, then both `adapter_id` and the version are required.  In our typical notebooks, we first prompt the base model (i.e., no adapter), then we prompt the fine-tuned adapter (in order to compare their performance).  Hence, the same wrapper function is being copied and pasted from one notebook to another:
```
def lorax_generate(
    client: Client,
    prompt: str,
    adapter_id: str | None = None,
    adapter_version: int | None = None,
    **kwargs,
) -> str:
    kwargs = kwargs or {}

    if adapter_id and adapter_version:
        adapter_id = f"{adapter_id}/{adapter_version}"

    response: Response = client.generate(
        prompt=prompt,
        adapter_id=adapter_id,
        **kwargs,
    )
    generated_text: str = response.generated_text

    return generated_text
```

This is repetitive and error prone.  Incorporating the flexibility and convenience inside LoRAX solves the issue.

## Before submitting
      to it if that's the case.
- [x] Did you write any new necessary tests?

## Who can review?

### DIFF
--- a/clients/python/lorax/client.py
+++ b/clients/python/lorax/client.py
@@ -143,6 +143,7 @@ class Client:
         self,
         prompt: str,
         adapter_id: Optional[str] = None,
+        adapter_version: Optional[int] = None,
         adapter_source: Optional[str] = None,
         merged_adapters: Optional[MergedAdapters] = None,
         api_token: Optional[str] = None,
@@ -173,8 +174,13 @@ class Client:
                 Input text
             adapter_id (`Optional[str]`):
                 Adapter ID to apply to the base model for the request
+            adapter_version: (`Optional[int]`):
+                Adapter version (applicable only when `adapter_source` is "pbase").
+                The format of `adapter_id` is "<ID>[/version]".  If `adapter_version` is
+                specified, then `adapter_id` is suffixed by `/adapter_version`, so for
+                convenience, `adapter_id` and `adapter_version` can be passed in separately.
             adapter_source (`Optional[str]`):
-                Source of the adapter (hub, local, s3)
+                Source of the adapter ("hub", "local", "s3", "pbase")
             merged_adapters (`Optional[MergedAdapters]`):
                 Merged adapters to apply to the base model for the request
             api_token (`Optional[str]`):
@@ -234,6 +240,7 @@ class Client:
         # Validate parameters
         parameters = Parameters(
             adapter_id=adapter_id,
+            adapter_version=adapter_version,
             adapter_source=adapter_source,
             merged_adapters=merged_adapters,
             api_token=api_token,
@@ -282,6 +289,7 @@ class Client:
         self,
         prompt: str,
         adapter_id: Optional[str] = None,
+        adapter_version: Optional[int] = None,
         adapter_source: Optional[str] = None,
         merged_adapters: Optional[MergedAdapters] = None,
         api_token: Optional[str] = None,
@@ -309,8 +317,13 @@ class Client:
                 Input text
             adapter_id (`Optional[str]`):
                 Adapter ID to apply to the base model for the request
+            adapter_version: (`Optional[int]`):
+                Adapter version (applicable only when `adapter_source` is "pbase").
+                The format of `adapter_id` is "<ID>[/version]".  If `adapter_version` is
+                specified, then `adapter_id` is suffixed by `/adapter_version`, so for
+                convenience, `adapter_id` and `adapter_version` can be passed in separately.
             adapter_source (`Optional[str]`):
-                Source of the adapter (hub, local, s3)
+                Source of the adapter ("hub", "local", "s3", "pbase")
             merged_adapters (`Optional[MergedAdapters]`):
                 Merged adapters to apply to the base model for the request
             api_token (`Optional[str]`):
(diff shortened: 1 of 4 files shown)


jeffreyftang (changes requested):
I'm not sure I understand the need for this PR. We can already pass the `myrepo/123` as the adapter_id.

tgaddair (commented):
Agree with @jeffreyftang, I would rather not couple the API to pbase versioning in this way.

alexsherstinsky:
> I'm not sure I understand the need for this PR. We can already pass the `myrepo/123` as the adapter_id.

@jeffreyftang The issue arises during the user experience.  In a notebook, we typically first prompt the base model, and then the fine-tuned adapter.  When we prompt the base model, the adapter/version parameter is left empty.  When we prompt the fine-tuned adapter, the user has to compose `myrepo/123` string.  This is inconvenient (cognitive mismatch) from the user standpoint, because the user operates at the level of abstraction of "calling LoRAX with parameters", whereas formatting a string is at a lower level.  For this reason, I have been copying the above convenience function (`def lorax_generate()`) across notebooks so that I only call this function with the appropriate arguments, without worrying about formatting strings.

So we have three choices:
0) Do nothing different, and we keep copying the convenience function over and over (less preferred from my perspective);
1) Allow the above convenience function to be part of the notebook utilities PR (in Predibase) -- seems acceptable to me.
2) Accept the proposed change in this PR (and also enhance it to handle the upcoming checkpointNum).

What do you think?  Thank you! /cc @tgaddair

jeffreyftang:
> When we prompt the base model, the adapter/version parameter is left empty. When we prompt the fine-tuned adapter, the user has to compose myrepo/123 string. This is inconvenient (cognitive mismatch) from the user standpoint, because the user operates at the level of abstraction of "calling LoRAX with parameters", whereas formatting a string is at a lower level.

I don't really understand the "formatting a string is at a lower level" objection. That's like saying the user can't be expected to construct the deployment URL as part of initializing the LoRAX client because it's string manipulation. The adapter_id is a string parameter, so constructing that string is just part of setting up the parameters to call LoRAX with.

> Do nothing different, and we keep copying the convenience function over and over.

It seems to me that this convenience function boils down to a personal preference. That's totally fine, but personal preferences don't belong in the base library.

Adding this new parameter makes the LoRAX client API more confusing (adapter version is a Predibase-specific concept) and gives it a larger error surface ("oh, I set adapter version but the adapter source is not `pbase`). Not wanting to do a simple `adapter_id=f"{name}/{version}"` does not seem to me a sufficient value-add to justify the change.

alexsherstinsky:
> > When we prompt the base model, the adapter/version parameter is left empty. When we prompt the fine-tuned adapter, the user has to compose myrepo/123 string. This is inconvenient (cognitive mismatch) from the user standpoint, because the user operates at the level of abstraction of "calling LoRAX with parameters", whereas formatting a string is at a lower level.
> 
> I don't really understand the "formatting a string is at a lower level" objection. That's like saying the user can't be expected to construct the deployment URL as part of initializing the LoRAX client because it's string manipulation. The adapter_id is a string parameter, so constructing that string is just part of setting up the parameters to call LoRAX with.
> 
> > Do nothing different, and we keep copying the convenience function over and over.
> 
> It seems to me that this convenience function boils down to a personal preference. That's totally fine, but personal preferences don't belong in the base library.
> 
> Adding this new parameter makes the LoRAX client API more confusing (adapter version is a Predibase-specific concept) and gives it a larger error surface ("oh, I set adapter version but the adapter source is not `pbase`). Not wanting to do a simple `adapter_id=f"{name}/{version}"` does not seem to me a sufficient value-add to justify the change.
@jeffreyftang To me the motivation was "what can be done to reduce the chance that the user will have a difficulty or make an error/typo, and give up trying to make Predibase work?"  For this reason, I felt that just allowing arguments would smooth the workflow.

@jeffreyftang I agree with you last argument that this makes it too `pbase`-specific.  Having said that, we do have in LoRAX `pbase` as one of the elements in the `ADAPTER_SOURCES` constant.  Regardless, I am currently exploring a notebooks-only option, which would remove or at least defer modifying the library.

alexsherstinsky:
Closing for now, since a "good-enough" solution for the convenience of supplying the adapter version as an argument to a function, rather than having to format a string was developed entirely at the notebook level (both Jupyter  and Colab).  Thanks to @jeffreyftang and @tgaddair for the discussion around this topic.

alexsherstinsky:
Closing for now, since a "good-enough" solution for the convenience of supplying the adapter version as an argument to a function, rather than having to format a string was developed entirely at the notebook level (both Jupyter  and Colab).  Thanks to @jeffreyftang and @tgaddair for the discussion around this topic.

PR #586, opened 2024-08-23
Add Llava Next (VLM)

tgaddair:


router/src/config.rs
```diff
@@ -0,0 +1,222 @@
...
+pub struct TextConfig {}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub struct VisionConfig {
+    pub(crate) image_size: usize,
+    pub(crate) patch_size: usize,
+}
+
+#[cfg(test)]
```
arnavgarg1, line 188: I'm assuming we plan to remove the test from here?

tgaddair: Wasn't planning on it. Why?

arnavgarg1: Oh I see. Wasn't sure if we have a separate test suite for our router or something, but if not, here is good!

PR #588, opened 2024-08-26
add new agnostic health endpoint

magdyksaleh:
Health endpoint that works across different model types. 

Main changes: 
1. Removed unused embedding model flag 
2. Added new fields in the shard info struct to reflect whether a model support classification, embedding or generation
3. Use those fields to run the correct health check when the endpoint is pinged 
4. Updated the flash attention function to support no kv cache for the embedding and classification models. (This should be fixed properly in follow up @tgaddair)

### DIFF
--- a/router/src/server.rs
+++ b/router/src/server.rs
@@ -1,6 +1,6 @@
 /// HTTP Server logic
 use crate::adapter::{extract_adapter_params, BASE_MODEL_ADAPTER_ID};
-use crate::health::Health;
+use crate::health::{self, Health};
 use crate::infer::{InferError, InferResponse, InferStreamResponse};
 use crate::validation::ValidationError;
 use crate::{json, HubTokenizerConfig};
@@ -82,16 +82,6 @@ async fn compat_generate(
 ) -> Result<Response, (StatusCode, Json<ErrorResponse>)> {
     let mut req = req.0;
 
-    if info.embedding_model {
-        metrics::increment_counter!("lorax_request_failure", "err" => "bad_request");
-        tracing::error!("Embedding model doesn't support generation");
-        let err = ErrorResponse {
-            error: "Embedding model doesn't support generation".to_string(),
-            error_type: "bad_request".to_string(),
-        };
-        return Err((StatusCode::BAD_REQUEST, Json(err)));
-    }
-
     // default return_full_text given the pipeline_tag
     if req.parameters.return_full_text.is_none() {
         req.parameters.return_full_text = Some(default_return_full_text.0)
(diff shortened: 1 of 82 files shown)


server/lorax_server/models/custom_modeling/flash_bert_modeling.py
```diff
@@ -164,7 +164,7 @@ def forward(self, hidden_states, cu_seqlens, max_s):
         q, k, v = qkv.view(-1, self.num_heads * 3, self.head_size).split(self.num_heads, dim=1)
 
         attn_output = torch.empty_like(q)
-        attention(q, k, v, attn_output, cu_seqlens, max_s, self.softmax_scale)
+        attention(q, k, v, None, None, cu_seqlens, max_s, self.softmax_scale)
```
tgaddair, line 167: Did you verify this works? You're not modifying the `attn_output`, which is now returned from the attention function, so I suspect this will effectively be a no-op.

magdyksaleh: ah man good catch!

magdyksaleh: fixed - good catch

server/lorax_server/utils/flash_attn.py
```diff
@@ -118,19 +119,29 @@ def attention(
...
+                q,
+                k,
+                v,
+                causal=causal,
+                window_left=window_size_left,
+                logits_soft_cap=softcap,
+                sm_scale=softmax_scale,
+            )
+
+        assert window_size_left == -1, "Windowing is not supported with flash infer when using kv cache"
```
tgaddair, line 144: We should keep this assertion at the top.

router/src/server.rs
```diff
@@ -374,19 +344,93 @@ responses(
...
-        )),
+async fn health(
+    infer: Extension<Infer>,
+    info: Extension<Info>,
+    request_logger_sender: Extension<Arc<mpsc::Sender<(i64, String, String)>>>,
+    req_headers: HeaderMap,
+    mut client: Extension<ShardedClient>,
+    mut health: Extension<Health>,
+) -> Result<(), (StatusCode, Json<ErrorResponse>)> {
+    println!("{:?}", health.shard_info());
```
tgaddair, line 356: Can remove debug print.

router/src/server.rs
```diff
@@ -374,19 +344,93 @@ responses(
...
-    match health.check().await {
-        true => Ok(()),
-        false => Err((
-            StatusCode::SERVICE_UNAVAILABLE,
-            Json(ErrorResponse {
-                error: "unhealthy".to_string(),
-                error_type: "healthcheck".to_string(),
-            }),
-        )),
+async fn health(
```
tgaddair, line 348: I wonder if we can just have this function return true if the metrics indicate the server has successfully completed a request in the last N seconds. I worry in high throughput scenarios that hitting this repeatedly could slow things down. But maybe we can leave that as a TODO.


cc @noyoshi

magdyksaleh: This was how the old endpoint was implemented. We can probably just store a bool and do lightweight checks. See: https://github.com/predibase/lorax/blob/main/router/src/health.rs#L25-L75

tgaddair: I like the use of the `generation_health` var, but would want to timestamp it so we re-run after some amount of time.

Fine to merge it as-is for now, but would like to revisit this soon as I do feel it's going to slow things down in the current form.

tgaddair (approved):

PR #593, opened 2024-08-27
support bge-base-en-v1.5

magdyksaleh:


arnavgarg1 (approved):

server/lorax_server/models/flash_bert.py
```diff
@@ -97,7 +97,9 @@ def __init__(
             dtype,
             process_group=self.process_group,
         )
-        prefix = None if (model_id == "WhereIsAI/UAE-Large-V1") else "bert"
+        prefix = "bert"
```
tgaddair, line 100: Is there a way to generalize this? For example, if the model is not a token classification model, or maybe we check the weight names, etc.?

PR #594, opened 2024-08-31
Fix bert ner

magdyksaleh:
Use the transformers implementation for now.

### DIFF
--- a/server/lorax_server/models/flash_bert.py
+++ b/server/lorax_server/models/flash_bert.py
@@ -97,7 +97,9 @@ class FlashBert(Model):
             dtype,
             process_group=self.process_group,
         )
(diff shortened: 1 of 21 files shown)


server/lorax_server/models/flash_bert.py
```diff
@@ -101,7 +103,7 @@ def __init__(
         if model_id in ["WhereIsAI/UAE-Large-V1", "BAAI/bge-base-en-v1.5"]:
             prefix = None
         if classifcation_head:
-            model = FlashBertModelForClassification(prefix, weights, device, dtype, config)
+            model = BertForTokenClassification.from_pretrained(model_id).to(device)
```
magdyksaleh, line 106: this is the main change

PR #608, opened 2024-09-17
Fix classify and classify_batch for Python client

tgaddair:


router/src/server.rs
```diff
@@ -1476,7 +1478,7 @@ async fn embed(
 async fn classify(
     infer: Extension<Infer>,
     Json(req): Json<ClassifyRequest>,
-) -> Result<(HeaderMap, Json<Vec<Entity>>), (StatusCode, Json<ErrorResponse>)> {
+) -> Result<(HeaderMap, Json<ClassifyResponse>), (StatusCode, Json<ErrorResponse>)> {
```
magdyksaleh, line 1481: Why are you changing the return type of this function? This nests the results under an 'entities' key?

tgaddair: This is to match the format expected by the client's `classify` function. We can change the client as well, but at the moment the contract is inconsistent between the two. In general, I prefer returning schemaed entities rather than un-schemed lists and such, though I understand why we're doing it for the batch call. Maybe the right change is to change the client in this instance so we don't break convention with the batch API call.

magdyksaleh: Agree

magdyksaleh (approved):

PR #614, opened 2024-09-18
Fix deps4

magdyksaleh:
kill all dependabot alerts


.github/workflows/build.yaml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - 'main'
+      - 'fix-deps4'
```
tgaddair, line 8: Please revert before landing.

tgaddair (approved):

PR #620, opened 2024-09-30
Create sync_from_s3.sh to fetch weights from S3 cache

joseph-predibase:
Adds a new script, `sync_from_s3.sh`. This script is a much simplified version of the existing `sync.sh` (which has been removed). This script attempts to fetch model weights from an S3 cache rather than Hugging Face directly. This makes Lorax more resilient to HF outages.

Also deletes container-entrypoint.sh, which was unused.

Tested by kicking off a Lorax container using instaging (staging agent has already been updated to enable the sync functionality if available).
* In Lorax builds with a sync script and a model from `llms.yaml`, this correctly fetches the weights from S3 (and lorax uses them, doesn't redownload).
* In Lorax builds with a sync script and a model not in `llms.yaml`, the script does not fetch weights from S3 (they're not there), but Lorax correctly fetches them from HF
* In Lorax builds without a sync script, the script is not run and Lorax fetches from HF.

I also compared the download speed of fetching from S3 vs HF. Downloading a 70b model from HF directly took 2 minutes, but from S3 (using s5cmd) took 1m45s. Obviously there's going to be some randomness so the improved speed may be a fluke - the important thing is that the numbers are roughly comparable.


launcher/src/main.rs
```diff
@@ -972,7 +973,10 @@ fn download_convert_model(
     loop {
         if let Some(status) = download_process.try_wait().unwrap() {
             if status.success() {
-                tracing::info!("Successfully downloaded weights.");
+                tracing::info!(
+                    "Successfully downloaded weights in {}s.",
+                    download_start.elapsed().as_secs()
```
joseph-predibase, line 978: Added download time.

Dockerfile
```diff
@@ -54,7 +54,8 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-ins
     ca-certificates \
     ccache \
     curl \
-    git && \
+    git \
+    less && \
```
joseph-predibase, line 58: One of the aws commands needs less for some reason.

gyanesh-mishra (approved):
Sync script changes look good to me, whether this should be in lorax or not, I'll leave it to @noyoshi @tgaddair to have strong takes on.

joseph-predibase:
> Sync script changes look good to me, whether this should be in lorax or not, I'll leave it to @noyoshi @tgaddair to have strong takes on.

Yeah, I'm writing up a little PR on the side to build a standalone container containing the sync script, so we can update the init container to use that container and not depend the Lorax version of a deployment. But right now the agent builds the init containers using a lorax image expecting this script so I figured I would build it so we can use the cache until the next release.

OTOH I'm gonna finish building the standalone container (good excuse to learn how to build a Docker container) and update the init containers to use it for the next release so maybe it's not worth merging in Lorax and we should just wait until the next release. Since this is in Lorax getting deployments to use it means a Lorax version bump - easy for serverless but a pain for dedicated customers.

joseph-predibase:
Abandoning this approach, we'll just use a dedicated image instead: https://github.com/predibase/predibase/pull/8675

PR #644, opened 2024-10-16
Enhance Structured Output Interface

### LINKED ISSUE #643: Enhance LoRAX `/v1/chat/completions` API to Align with OpenAI Structured Output Interface
### Feature request

I propose an enhancement to the `/v1/chat/completions` API in LoRAX to make it fully compatible with OpenAI's structured output interface. Currently, LoRAX's `response_format` only supports `type: "json_object"` and requires the `schema` field. However, OpenAI's interface allows for three types: `text`, `json_object`, and `json_schema`. Additionally, OpenAI uses the `json_schema` field instead of `schema`, and the `schema` field is not mandatory.

The proposal is to update the LoRAX's OpenAI-compatible API to match OpenAI's interface, allowing the use of all three types (`text`, `json_object`, `json_schema`), and `json_schema` optional. This would allow users to generate arbitrary JSON objects without requiring a predefined schema, enhancing compatibility and flexibility when using structured outputs.

For reference, here are the relevant OpenAI documentation links:
- https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format
- https://platform.openai.com/docs/guides/structured-outputs

### Motivation

Currently, LoRAX differs from OpenAI's structured output interface in that it only supports `json_object` with a mandatory `schema` field. This causes frustration when trying to use the API in OpenAI-compatible environments, as OpenAI allows for more flexibility (including text and structured JSON outputs). Aligning LoRAX with OpenAI's behavior would greatly enhance compatibility and make it easier for developers who work with both systems.

This change would be especially useful for developers migrating between OpenAI and LoRAX or those integrating LoRAX into OpenAI-compatible applications. Additionally, this could potentially avoid confusion for users expecting OpenAI-like behavior.

### Your contribution

I would be happy to contribute by submitting a PR that implements the changes required to make `response_format.schema` optional and allows arbitrary JSON object generation, while retaining the original LoRAX interface functionality. If the proposal aligns with the maintainers' vision, I can submit this enhancement for review.

Thank you for considering this request!

GirinMan:
# What does this PR do?
This PR enhances the `/v1/chat/completions` API in LoRAX to improve compatibility with OpenAI’s structured output interface. The changes include:

- Adding support for `text`, `json_object`, and `json_schema` as valid `type` options in `response_format`.
- Making the `schema` (or `json_schema`) field optional, allowing the generation of arbitrary JSON objects when using `json_object` without requiring a predefined schema.
- Retaining the original LoRAX interface behavior for backward compatibility.

This update brings LoRAX's API closer in line with OpenAI’s, allowing developers to more easily switch between or integrate both systems.

Fixes #643 

## Before submitting
- [x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x] Was this discussed/approved via a Github issue or the discord / slack channel? Please add a link
      to it if that's the case.

## Who can review?

@tgaddair
@donggyukimc


tgaddair (approved):
LGTM! I think the one thing I might do as a follow-up is allow specifying a JSON schema even when the format is `json_object`, that would ensure backwards compatibility for existing LoRAX users. Do you have any objections to making that change?

GirinMan:
> LGTM! I think the one thing I might do as a follow-up is allow specifying a JSON schema even when the format is `json_object`, that would ensure backwards compatibility for existing LoRAX users. Do you have any objections to making that change?

That sounds good to me too. Because it's not a big deal, I'll make a new PR handling it in few days.

PR #652, opened 2024-10-17
Support FP8 KV Cache

ajtejankar:
Tested GLUE tasks with and without adapter and the accuracy is as expected. Previous implementation using vllm kernels didn't work in this setting. Uses static scales obtained with ultrachat 2k. The cache is stored in E4M3 format. Test weights at `ajinkya-tejankar/Mistral-7B-Instruct-v0.2-FP8-UltraChat-2000-KV`.


server/lorax_server/utils/paged_attention.py
```diff
@@ -33,17 +36,22 @@ def reshape_and_cache(
     key_cache: torch.Tensor,
     value_cache: torch.Tensor,
     slots: torch.Tensor,
+    k_scale: float = 1.0,
+    v_scale: float = 1.0,
 ):
     if FLASH_INFER:
+        if key_cache.dtype == torch.float8_e4m3fn and value_cache.dtype == torch.float8_e4m3fn:
```
ajtejankar, line 43: @tgaddair This is a potential bottleneck, so I am trying to search for a fused kernel.

server/lorax_server/models/custom_modeling/flash_mistral_modeling.py
```diff
@@ -205,7 +205,7 @@ def _load_gqa(config, prefix: str, weights, head_size):
     if type(weight) is tuple:
         weight, input_scale, weight_scale = weight
 
-    if config.quantize not in ["gptq", "awq", "fp8"]:
+    if config.quantize not in ["gptq", "awq", "fp8", "fp8_kv"]:
```
tgaddair, line 208: can you make this a utility function so we don't need to reimplement for every model?

server/lorax_server/models/custom_modeling/flash_mistral_modeling.py
```diff
@@ -260,6 +260,14 @@ def __init__(
             )
         self.num_heads = self.num_heads // weights.process_group.size()
         self.num_key_value_heads = config.num_key_value_heads // weights.process_group.size()
+        if paged_attention.is_fp8_supported() and config.quantize and config.quantize.endswith('_kv'):
```
tgaddair, line 263: This is a little hacky. Can we have a function like `is_fp8_kv` or something instead of checking for a suffix?

server/lorax_server/models/custom_modeling/flash_mistral_modeling.py
```diff
@@ -342,12 +358,14 @@ def forward(
                 query,
                 torch.select(kv, dim=1, index=0),
                 torch.select(kv, dim=1, index=1),
-                kv_cache[0],
-                kv_cache[1],
+                None if self.kv_dtype == 'fp8' else kv_cache[0],
```
tgaddair, line 361: I would rather pass in the `kv_cache` values and then have this function decide whether to use them or not (as this seems like a low level implementation detail). One way would be to have a param you pass in here like `is_fp8_kv`.

server/lorax_server/models/flash_causal_lm.py
```diff
@@ -956,6 +956,13 @@ def __init__(
         config = config_cls.from_pretrained(model_id, revision=revision, trust_remote_code=trust_remote_code)
         config.quantize = quantize
 
+        from lorax_server.utils import paged_attention
```
tgaddair, line 959: Can we import at the top?

server/lorax_server/models/flash_causal_lm.py
```diff
@@ -956,6 +956,13 @@ def __init__(
...
+        if paged_attention.is_fp8_supported() and config.quantize and config.quantize.endswith('_kv'):
```
tgaddair, line 960: Should be a utility function, since same code is used elsewhere in the mistral modeling.

server/lorax_server/utils/torch_utils.py
```diff
@@ -13,7 +13,7 @@ def is_bf16_supported() -> bool:
 def is_fp8_quantized(config, layer_name):
     # check if quantization is fp8 and either of the fused layers is not ignored
     # typically, either all qkv will be quantized or none so just check for one
-    if config.quantize == "fp8" and hasattr(config, "quantization_config"):
+    if config.quantize and config.quantize.startswith("fp8") and hasattr(config, "quantization_config"):
```
tgaddair, line 16: I would also make this a utility function. Reason being: we might not continue to adhere to this naming convention in the future.

ajtejankar: This is actually dead code. Removed it.

server/lorax_server/models/flash_causal_lm.py
ajtejankar: Yes, moved the import to the top. Server tests were breaking with the error `could not import vllm` when at top so had it here, but now that `is_fp8_supported` is in `torch_utils` even this shouldn't happen.

server/lorax_server/models/custom_modeling/flash_mistral_modeling.py
ajtejankar: Done.

server/lorax_server/models/custom_modeling/flash_mistral_modeling.py
ajtejankar: Done.

server/lorax_server/models/custom_modeling/flash_mistral_modeling.py
ajtejankar: Done.

tgaddair (approved):
LGTM!

samagra14:
Suddenly started getting these errors - `'FlashLlamaAttention' object has no attribute 'fp8_kv'` since about a few hours ago both on L40S and A10Gs. 

I am guessing it is because of this merge. Any idea how to resolve ?

ajtejankar:
Should be fixed with #662

PR #653, opened 2024-10-17
Chunked prefill

tgaddair:


magdyksaleh (approved):
Looks good. I want to add a test to make sure this doesn't break embedding and classification models but I can follow up with a pr

ajtejankar:
Looks good to me as well. I'll test it out once it's merged in.

tgaddair:
Good callout @magdyksaleh! I did test embeddings (though not NER) and got it working :)

PR #621, opened 2024-09-30
support MRL embeddings for qwen2

magdyksaleh:

PR #654, opened 2024-10-22
Fix PREDIBASE_API_TOKEN env var being thrown away

joseph-predibase:
Fixes a regression introduced in #621. The Lorax launcher sets the environment variable PREDIBASE_API_TOKEN from the CLI flag --predibase-api-token. Unfortunately that PR overwrites the env var struct after the PREDIBASE_API_TOKEN has been set. :-(

Two deployments use this argument, 4421 and 4781. The latter uses an image that was build yesterday and so includes this bug, we will need to udpate it once this PR merges. (4421 is using an image that was built before this bug).

### DIFF
--- a/launcher/src/main.rs
+++ b/launcher/src/main.rs
@@ -522,9 +522,6 @@ fn shard_manager(
         fs::remove_file(uds).unwrap();
     }
 
-    // Copy current process env
-    let mut envs: Vec<(OsString, OsString)> = env::vars_os().collect();
-
     // Process args
     let mut shard_args = vec![
         "serve".to_string(),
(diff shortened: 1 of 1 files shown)


launcher/src/main.rs
```diff
@@ -621,6 +611,13 @@ fn shard_manager(
     // Copy current process env
     let mut envs: Vec<(OsString, OsString)> = env::vars_os().collect();
```
joseph-predibase, line 612: This line was introduced in #621 , which means that the value of `envs` is overwritten and no longer contains the value of `PREDIBASE_API_TOKEN`.

magdyksaleh (approved):

PR #655, opened 2024-10-22
Prompt prefix caching for multi-LoRA

tgaddair:


.github/workflows/build.yaml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - 'main'
+      - 'lora-prompt-cache'
```
tgaddair, line 8: TODO: remove before landing

ajtejankar (approved):

PR #663, opened 2024-11-01
added metrics docs, updated links in main docs

noyoshi:


docs/reference/metrics.md
```diff
@@ -0,0 +1,19 @@
+# Metrics
```
tgaddair, line 1: Need to add this to `mkdocs.yml` so it shows up in nav.

tgaddair (approved):

PR #666, opened 2024-11-06
Fix sliding window + compile bug

ajtejankar:
The `clamp` method on `Seqlen` created a new tensor which CUDA graphs don't like. So, modified it to clamp in place.


arnavgarg1 (approved):
LGTM

PR #668, opened 2024-11-08
Fix stella embeddings + Integration tests for lorax

magdyksaleh:


.github/workflows/build.yaml
```diff
@@ -1,15 +1,16 @@
-name: Build and push docker image to ghcr
+name: Build and push docker image to ghcr and run integration tests
 
 on:
   workflow_dispatch:
   push:
     branches:
       - 'main'
+      - 'integration-tests-main' # XXX: This is a temporary branch to run integration tests on the main branch
```
tgaddair, line 8: Remove before landing.

.github/workflows/build.yaml
```diff
@@ -103,10 +104,33 @@ jobs:
           file: ./Dockerfile  # Path to your Dockerfile
           push: false
           tags: ${{ steps.meta.outputs.tags }}
+          load: true
           outputs: type=oci,compression=gzip,dest=${{ steps.vars.outputs.image_path }}-${{ steps.vars.outputs.tag_hash }}.tar.gz
           cache-from: type=local,src=${{ steps.vars.outputs.cache_dir }}
           cache-to: type=local,mode=max,image-manifest=true,oci-mediatypes=true,dest=${{ steps.vars.outputs.cache_dir }}
 
+      # # Keeping this here since it's really useful when debugging this gha
```
tgaddair, line 112: Maybe make this comment more clear, like: "Uncomment to ... when debugging integration tests", etc.

integration-tests-2/test_base_llms.py
```diff
@@ -0,0 +1,47 @@
+import requests
+from utils.docker_runner import run_lorax_container
+
+
+def test_base_mistral():
```
tgaddair, line 5: This will be fun to make data-driven using `pytest.mark.parametrize` in a follow-up :)

But good for v1.

integration-tests-2/utils/docker_runner.py
```diff
@@ -0,0 +1,76 @@
...
+    client = _init_client()
+    container = _start_container(client, config)
+    print("Started container")
+    try:
+        print("Waiting for container to be healthy")
+        _wait_for_healthy(container)
+        print("Container is healthy")
+        yield
+    finally:
+        print("Stopping container")
```
tgaddair, line 74: Would suggest using logging instead of prints to make it easier to filter it out.

tgaddair (approved):
LGTM! Why the name `integration-tests-2`, maybe we just need to rename `integration-tests` to something more meaningful? Can be a follow-up.

PR #669, opened 2024-11-08
Add support for adapter loading in mllama

ajtejankar:
Supporting adapter loading in VLMs, specifically mllama models. This is more complex than we had initially anticipated as a VLM is under the hood a composite model with multiple trunks which requires separating and loading the lora weights independently for each trunk. Further, currently we don't have lora adapters for certain layers in one of the trunks (cross attention layers) which requires careful handling. Finally, the linear layer classes currently used don't support adapter weights (currently investigating).

### DIFF
--- a/server/lorax_server/adapters/lora.py
+++ b/server/lorax_server/adapters/lora.py
@@ -152,11 +152,19 @@ class LoraWeights(AdapterWeights):
         layer_type: str,
         unused_weight_names: Set[str],
     ) -> Optional[AdapterWeights]:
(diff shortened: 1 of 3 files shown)


ajtejankar:
The PR is ready from my side. Loading randomized adapter results in garbage output while the model answers correctly without it. Slowly changing the noise in the randomized adapter also slowly degrades the quality of the model's output again indicating that adapters are being loaded correctly.

tgaddair (approved):
LGTM!

PR #677, opened 2024-11-12
Only push main if int tests pass

magdyksaleh:


.github/workflows/build.yaml
```diff
@@ -124,6 +124,16 @@ jobs:
       #       docker tag ghcr.io/predibase/lorax:main "$tag"
       #     done <<< "$tags"
 
+      - name: Run integration tests
```
tgaddair, line 127: Ideally I would still want to push the image before integration tests run. This is important in cases where we need to get an image out for an urgent issue, etc.

What I would suggest: we have two separate actions for pushing main and pushing all other tags. The first action runs before int tests, the second runs after int tests. What do you think?

magdyksaleh: agree!

tgaddair (approved):
LGTM!

PR #680, opened 2024-11-13
Format fixes tool calling

magdyksaleh:
Allow content to be null and to accept parameters as dicts and arguments as strings for function calling


jeffreyftang (approved):

tgaddair (approved):
LGTM!

PR #691, opened 2024-11-21
Handle case where adapter exists but is still being trained

tgaddair:


server/lorax_server/models/mllama.py
jeffreyftang, line 1: I think this unintentionally got committed?

server/lorax_server/models/flash_causal_lm.py
jeffreyftang, line 1: I think this unintentionally got committed?

tgaddair: This was from the formatter.

jeffreyftang (approved):

PR #693, opened 2024-11-22
Add support for targeting cross_attn layers in mllama

ajtejankar:
Previous PR that added support for mllama models didn't allow targeting cross attention layers in the text model. This PR adds support for it. I think the code is ready, but it needs to be tested with an adapter.


ajtejankar:
I've verified this against an adapter targeting all modules that was trained on a small subset of the [Doc VQA dataset](https://huggingface.co/datasets/cmarkea/doc-vqa). It's ready to merge.

arnavgarg1 (approved):
LGTM! Let's merge after resolving the merge conflict

PR #702, opened 2024-12-03
Openai embedding endpoint

magdyksaleh:
Adding an OpenAI compatible embedding endpoint. 

Also adding a batch embed functionality

### DIFF
--- a/router/src/server.rs
+++ b/router/src/server.rs
@@ -1484,6 +1484,7 @@ pub async fn run(
         .route("/generate_stream", post(generate_stream))
         .route("/v1/completions", post(completions_v1))
         .route("/v1/chat/completions", post(chat_completions_v1))
+        .route("/v1/embeddings", post(embed))
         // AWS Sagemaker route
         .route("/invocations", post(compat_generate));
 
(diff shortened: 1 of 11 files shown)


tgaddair (approved):
Sweet! LGTM!

router/src/server.rs
```diff
@@ -1643,6 +1645,69 @@ async fn embed(
...
+)]
+#[instrument(skip_all)]
+#[axum::debug_handler]
+async fn compat_embed(
+    infer: Extension<Infer>,
+    Json(req): Json<CompatEmbedRequest>,
+) -> Result<Json<CompatEmbedResponse>, (StatusCode, Json<ErrorResponse>)> {
+    metrics::increment_counter!("lorax_request_count");
+    tracing::debug!("Input: {}", req.input);
+    if let StringOrVec::Vec(inputs) = req.input {
```
arnavgarg1, line 1667: OOC what's the difference between string or vec as inputs? Is it the vec just a list of strings and why we have a .map call?

magdyksaleh: yeah but we need to handle the case that the input is a string or a list of strings. Rust is a bit weird like that. So you need a special type for that. https://platform.openai.com/docs/api-reference/embeddings/create#embeddings-create-input

arnavgarg1: Very interesting, good to know!

PR #705, opened 2024-12-05
Fix decoder_input_details bug

ajtejankar:
Replace `PrefillTokens` with `NextTokens`

### DIFF
--- a/proto/generate.proto
+++ b/proto/generate.proto
@@ -214,15 +214,6 @@ message GeneratedText {
     optional uint64 seed = 5;
 }
 
-message PrefillTokens {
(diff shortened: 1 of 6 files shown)


proto/generate.proto
```diff
@@ -249,7 +249,7 @@ message Generation {
     /// Request ID
     uint64 request_id = 1;
     /// Prefill tokens (optional)
-    PrefillTokens prefill_tokens = 2;
+    NextTokens prefill_tokens = 2;
```
tgaddair, line 252: Should we just remove the `PrefillTokens` proto from here if it's not used anywhere?

PR #714, opened 2024-12-11
some metrics and logging for batch classify

magdyksaleh:


arnavgarg1 (approved):

router/src/server.rs
```diff
@@ -1858,6 +1860,8 @@ async fn classify_batch(
     );
 
     let batch_entity_vec = responses.into_iter().map(|r| r.predictions).collect();
+    tracing::debug!("Output: {:?}", batch_entity_vec);
+    tracing::info!("Success");
```
joseph-predibase, line 1864: Let's add this to `classify()` while we're here.

joseph-predibase (commented):
These seem reasonable. I don't know Lorax well enough but it feels like we should be able to add more than this. I don't know what else we could add, but when we looked we basically just had request count, input size, and duration. Seems like we should at the very least be able to add some more metrics on the output.

PR #715, opened 2024-12-12
Fix FlashInfer + Medusa bug

ajtejankar:
With vLLM's paged attention kernels `block_tables` is handled correctly and there's no other tensor used to index in this tensor. However, FlashInfer requires another tensor to index into `block_tables` (this is called as `indices` in the code) called `indptr`. For speculative tokens, we flatten the `block_tables` and treat the tokens as if they were a part of a different request altogether. The simplest fix for this problem is to make the behavior is FlashInfer similar to vLLM kernels. The current implementation is very hacky, and I am working on refining it.


tgaddair (approved):
LGTM!

PR #720, opened 2024-12-16
Memory estimate fixes

magdyksaleh:
- Update mem wiggle room to 0.9 (from 0.8)
- Ensure free_memory is never negative
- Lazy load graph for larger compile batch sizes
- Add new argument compile_batch_size (default 32) which is the batch size we will initially use for cuda compilation of models
- make the precommit use ruff instead of flake8

### DIFF
--- a/server/lorax_server/utils/graph.py
+++ b/server/lorax_server/utils/graph.py
@@ -41,7 +41,7 @@ BATCH_SIZE_INCREMENT = 32
 CACHED_BATCH_SIZES = [1, 2, 3, 4, 8, 16] + [
     BATCH_SIZE_INCREMENT * (i + 1) for i in range(MAX_BATCH_SIZE // BATCH_SIZE_INCREMENT)
 ]
-CACHED_BATCH_SIZES = [b for b in CACHED_BATCH_SIZES if b <= MAX_BATCH_SIZE]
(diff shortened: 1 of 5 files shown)


server/lorax_server/utils/graph.py
```diff
@@ -603,7 +605,13 @@ def forward(
 
         key = (batch_size, max_rank)
         graph = self.cache.get(key)
-        if graph is None or not graph.input_state.traced_adapter_layer_names.issuperset(adapter_data.layer_names()):
+        if (
+            graph is None
+            or not graph.input_state.traced_adapter_layer_names.issuperset(adapter_data.layer_names())
+            # This is the case where COMPILE_BATCH_SIZE < batch_size <= MAX_BATCH_SIZE so
+            # we just retrace the graph for that new size
+            or batch_size > self.batch_size
```
magdyksaleh, line 613: https://github.com/predibase/lorax/blob/12e530a0be722871be1cfb3e791f9aed48f66dca/server/lorax_server/utils/graph.py#L478-L498

This ensures that we stay within the range

tgaddair (changes requested):
Actually, I think we need changes to warmup.

tgaddair (approved):
LGTM!

PR #732, opened 2025-01-08
Enable eager prefill by default

ajtejankar:


arnavgarg1 (approved):

arnavgarg1:
Can we update the default values in our docs as well to reflect this change?

launcher/src/main.rs
```diff
@@ -508,7 +508,7 @@ struct Args {
     /// Whether to use the prefix caching mechanism. This will skip computing attention on previously cached prefixes
     /// in the prompt. Useful in cases where many queries need to be run over a shared context, or for long multi-turn
     /// chats conversations.
-    #[clap(long, env)]
+    #[clap(default_value = "true", long, env)]
```
tgaddair, line 511: Are we confident this is stable enough to enable by default? I believe there are some open questions on this at the moment. Maybe we can start with just landing the eager prefill change.

ajtejankar:
@arnavgarg1 Not sure if this is the only thing that you meant, but I've added a line documentation to `launcher.md` for `--eager-prefill` and reverted `--prefix-caching` to be false by default.

PR #734, opened 2025-01-09
Add eager flag and set compile to be always true

ajtejankar:
Adds a new flag to the launcher called `--eager` that allows users to disable CUDA graph compilation which is enabled by default. Also, sets `--compile` to always be true. Finally, tested that compile works with following list of models, and added a property called `supports_cuda_graph_compilation` that individual models can set to disable compilation.

* microsoft/phi-4
* mistralai/Mistral-7B-Instruct-v0.2
* meta-llama/Llama-3.2-11B-Vision-Instruct
* meta-llama/Llama-3.1-8B-Instruct
* Qwen/Qwen2.5-7B-Instruct
* upstage/SOLAR-10.7B-v1.0
* ibm-granite/granite-3.1-8b-instruct

### DIFF
--- a/launcher/src/main.rs
+++ b/launcher/src/main.rs
@@ -753,8 +753,7 @@ fn shard_manager(
     }
 
     if compile && eager {
-        tracing::error!("Cannot use both --compile and --eager at the same time.");
-        return;
+        panic!("Cannot use both --compile and --eager at the same time.");
     }
 
     // Speculative decoding
--- a/launcher/src/main.rs
+++ b/launcher/src/main.rs
@@ -342,8 +342,17 @@ struct Args {
 
     /// Whether you want to compile the model into a CUDA graph.
     /// This will speed up decoding but increase GPU memory usage.
-    #[clap(long, env, value_enum)]
+    /// Only use either `--compile` or `--eager`. Using both at the same time will
+    /// result in an error.
+    #[clap(default_value = "true", long, env, value_enum)]
     compile: bool,
+
+    /// Whether you want to run the model in eager mode, without
+    /// CUDA mode compilation, or run it with compilation.
+    /// Only use either `--compile` or `--eager`. Using both at the same time will
+    /// result in an error.
+    #[clap(default_value = "false", long, env, value_enum)]
+    eager: bool,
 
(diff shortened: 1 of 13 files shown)


launcher/src/main.rs
```diff
@@ -342,9 +342,18 @@ struct Args {
 
     /// Whether you want to compile the model into a CUDA graph.
     /// This will speed up decoding but increase GPU memory usage.
-    #[clap(long, env, value_enum)]
+    /// Only use either `--compile` or `--eager`. Using both at the same time will
+    /// result in an error.
+    #[clap(default_value = "true", long, env, value_enum)]
```
tgaddair, line 347: Since this defaults to true here, will this raise an error if the user specifies `--eager` without setting anything for `compile`?

Typically we would change the type of `compile` here to `Option[bool]` so we can distinguish between explicitly specified args vs unspecified.

server/lorax_server/models/model.py
```diff
@@ -228,6 +228,10 @@ def check_initialized(self):
                 f"found uninitialized parameters in model {self.__class__.__name__}: {uninitialized_parameters}"
             )
 
+    @property
+    def supports_cuda_graph_compilation(self) -> bool:
+        return True
```
tgaddair, line 233: Do we need to set this to False for some models?

ajtejankar: At least with the limited models I tested, compile worked so we don't set this property to False for any model, but we may want to in the future. Not needed at the moment so happy to get rid of it.

launcher/src/main.rs
tgaddair: How do we detect the `--compile` is not provided?

server/lorax_server/models/model.py
tgaddair: Check for lines like this in the code:

```
if compile:
            raise ValueError("`--compile` is not supported with CausalLM")
```

There are a number of them

tgaddair (approved):
LGTM, though not clear how the function supports graph compilation is used. I would think it would be a check before we initialize the model and raise an error is compile explicitly set. But I think this is more a nice to have

PR #735, opened 2025-01-09
offline mode support

noyoshi:
supports offline mode for non embedding models.


.github/workflows/build.yaml
```diff
@@ -4,7 +4,8 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - 'main'
+      - "main"
+      - offline-mode-support
```
tgaddair, line 8: Please remove before landing.

PR #742, opened 2025-01-17
fix block size in health check

magdyksaleh:
# What does this PR do?

Fixes # (issue)

## Before submitting
      to it if that's the case.

## Who can review?

### DIFF
--- a/router/src/health.rs
+++ b/router/src/health.rs
@@ -37,6 +37,7 @@ impl Health {
     }
 
     pub(crate) async fn check_generation(&mut self) -> bool {
+        let block_size = self.shard_info().block_size;
         let generation_liveness_request = Request {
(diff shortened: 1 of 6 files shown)


router/src/health.rs
```diff
@@ -66,7 +66,7 @@ impl Health {
             adapter_index: 0,
             // Block 0 is reserved for health checks
             blocks: vec![0],
-            slots: (0..16).collect(),
+            slots: (0..self.shard_info().block_size).collect(),
```
tgaddair, line 69: Nit, but can we use the property directly instead of going through a function call that adds extra indirection of function call?

magdyksaleh: sure

tgaddair (approved):

PR #749, opened 2025-01-27
Set adapter memory fraction to 0 by default

ajtejankar:


arnavgarg1:
@ajtejankar This should only be set to 0.0 when we have preloaded adapter IDs, but not otherwise. Is that what's happening here?

magdyksaleh (approved):

PR #750, opened 2025-01-27
Add ability to specify revision in `model_id` flag using model@revision format

joseph-predibase:
Adds support to specify a model revision in the format `model@revision` as part of the `--model-id` flag. Mutually exclusive with the `--revision` flag.

Tested by deploying an LLM specifying a non-default revision and verified that we fetched the desired version from HF.


magdyksaleh (approved):
great!

PR #752, opened 2025-01-29
Improve structured output generation times

jeffkinnison:
# What does this PR do?

Recently, I did a benchmark of structured output generation with lorax vs vllm. The benchmark compared the per-request running time of prompting `meta-llama/Llama-3.2-1B-Instruct` to generate JSON using structured output and standard generation (`none`) with lorax, and using the `outlines`, `xgrammar`, and `lm-format-enchancer` backends in vllm. The plot below shows the results of timing 50 requests to each backend.

image

I found lorax structured output to be significantly slower than both standard prompting with lorax and structured output with vllm + outlines.

This PR addresses the slower inference times in two ways
- Updating `outlines` to 0.1.1 using [this TGI update](https://github.com/huggingface/text-generation-inference/pull/2742) as a guide
- Removing [out of bounds token filtering logic](https://github.com/predibase/lorax/blob/05b2d63845a6236727b25af22ac22a701f8913da/server/lorax_server/utils/logits_process.py#L556-L558) from `OutlinesLogitsProcessor`.

After making these changes, structured output with lorax takes ~0.4s per example in the same benchmark, and I have yet to encounter out of bounds tokens.

## Before submitting
      to it if that's the case.


tgaddair (approved):
Nice! LGTM.

PR #756, opened 2025-02-05
Fix memory calculation causing --compile to error

ajtejankar:
The exact reason of the error is not clear to me, but there's a simpler way to calculate the free memory which is implemented here. Here's what I've understood so far.
1. When compile is enabled, the code path that estimates cuda graph memory causes PyTorch to release temporary memory which increases the amount of free memory available
2. This doesn't happen when compile is not enabled which results in lesser free memory
3. My guess is that free memory is underestimated in without compile case but is slightly overestimated in with compile case
4. We need to do two things: align the memory estimation between two cases and correctly estimate the memory in both cases
5. We can align the memory estimation in both cases by adding a call to `empty_cache`
6. We can make the memory estimation accurate by getting rid of the batch kv cache by setting `self.kv_cache = []` before `empty_cache` call and by not adding `batch_num_blocks` to `num_blocks`.

### DIFF
--- a/server/lorax_server/models/flash_causal_lm.py
+++ b/server/lorax_server/models/flash_causal_lm.py
@@ -1344,6 +1344,8 @@ class FlashCausalLM(Model):
             logger.info("Estimated graph cache memory: {} MB", graph_cache_memory / 1024 / 1024)
             torch.cuda.synchronize(self.device)
 
+        self.kv_cache = []
+        torch.cuda.empty_cache()
         # Inspired by the original implementation in [vllm](https://github.com/vllm-project/vllm)
         # Calculate the number of blocks that can be allocated with the free memory
         dtype_size = torch.tensor([], dtype=self.dtype).element_size()
@@ -1358,13 +1360,7 @@ class FlashCausalLM(Model):
         free_memory = max(0, free_memory - graph_cache_memory)
         logger.info("Memory remaining for kv cache: {} MB", free_memory / 1024 / 1024)
 
-        batch_num_blocks = batch.num_blocks if batch is not None else 0
-        num_blocks = (
-            # Leave 5% for some wiggle room
-            int((free_memory * MEMORY_WIGGLE_ROOM) // total_cache_size)
-            # Add batch.num_blocks as we allocated it above, so it is included in the peak memory.
-            + batch_num_blocks
-        )
+        num_blocks = int((free_memory * MEMORY_WIGGLE_ROOM) // total_cache_size)
 
         del batch
 
--- a/server/lorax_server/models/flash_causal_lm.py
+++ b/server/lorax_server/models/flash_causal_lm.py
@@ -1291,7 +1291,7 @@ class FlashCausalLM(Model):
         )
 
     def warmup(self, batch: FlashCausalLMBatch, max_new_tokens: int, embedding_model: bool = False):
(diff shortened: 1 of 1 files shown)


ajtejankar:
There's one more problem that the previous commit [03726e7](https://github.com/predibase/lorax/pull/756/commits/03726e7977a7dee1203bfa3c144d8e23fbdd5cb9) didn't fix. The `mdoel_graph_wrapper` object contains a reference to the `kv_cache` memory which prevents torch from freeing it up when `empty_cache` is called. Hence, we use `model_graph_wrapper` to estimate cuda graph memory overhead, delete it, and then re-instantiate one final time to actually warmup the graphs.

server/lorax_server/models/flash_causal_lm.py
```diff
@@ -1326,24 +1346,17 @@ def warmup(self, batch: FlashCausalLMBatch, max_new_tokens: int, embedding_model
...
-                self.layer_to_lora_weights,
-                self.punica_wrapper,
-            )
+            self.init_graph_wrapper(max_total_tokens)
             graph_cache_memory = self.model_graph_wrapper.get_estimated_cache_memory()
             logger.info("Estimated graph cache memory: {} MB", graph_cache_memory / 1024 / 1024)
             torch.cuda.synchronize(self.device)
 
+        logger.info(f'Post warmup cuda memory: {get_cuda_free_memory(self.device, 1) / (1024 ** 3)}')
+        del batch
```
tgaddair, line 1355: Hmm, I don't think you can delete the batch at this point, as it's needed to estimate the prefill memory overhead. I would compare with how TGI does it.

server/lorax_server/models/flash_causal_lm.py
```diff
@@ -1326,24 +1346,17 @@ def warmup(self, batch: FlashCausalLMBatch, max_new_tokens: int, embedding_model
...
+        del self.model_graph_wrapper
+        self.kv_cache = []
+        torch.cuda.empty_cache()
```
tgaddair, line 1358: You may want to add an additional `torch.cuda.synchronize()` just before this empty cache call to ensure that the delete op has been executed.

server/lorax_server/models/flash_causal_lm.py
```diff
@@ -1326,24 +1346,17 @@ def warmup(self, batch: FlashCausalLMBatch, max_new_tokens: int, embedding_model
...
+        logger.info(f'Post warmup empty_cache cuda memory: {get_cuda_free_memory(self.device, 1) / (1024 ** 3)}')
```
tgaddair, line 1359: Can you include units in this log message? Looks like GiB.

tgaddair (commented):
Nice, added a couple comments.

server/lorax_server/models/flash_causal_lm.py
ajtejankar: Okay they seem to delete the batch [after `num_block` calculation](https://github.com/huggingface/text-generation-inference/blob/856709d5c3d24a9cad53c2ab94835c6190bacf3c/server/text_generation_server/models/flash_causal_lm.py#L1609) as well, so I'll do the same.

ajtejankar: Basically, I'll revert back to whatever we had.

tgaddair (approved):
LGTM!

